Issue #4, opened 2024-03-04
Ideas on generalization of spatial package backends and file sources using GDAL (terra, sf, stars, etc.)


brownag:
I wanted to throw up some ideas for discussion, might be a bit rambling for a single issue. Happy to break off any particular items as new issues or address in specific PRs; I will submit some draft PRs once I have fleshed these ideas out. I say "we" a lot in here but ultimately I am just one interested opinion and welcome any thoughts or alternatives.

---

The [current](https://github.com/njtierney/geotargets/commit/52670f0d6b2cc50141f280d6859b0bfbd879a289) target storage format functions defined are file-format centric. This is great, because GDAL is the library behind the scenes for common interfaces to a variety of different file formats. GDAL is used in several R spatial packages notably: sf, terra, and stars. I think this project should abstract out the functionality for GDAL data source paths and provide support for multiple R package/object type interfaces in the result the user sees.

In my opinion, {geotargets} should provide default behavior based on type of spatial data, i.e. vector geometry vs. raster--this is so the user doesn't have to think too much about the formats in their target store, just that they are able to roundtrip an R object equivalent to what they started with. If they care about the format, they should have the ability to choose.

I'd like to make (or suggest others make) a couple PRs to implement:
 1) Spatial backend options to allow, for example: GeoTIFF format object with {stars} or Shapefile format with {sf}
 2) Generalization of the "multiple file target compression" GDAL /vsizip/ approach to all backends and formats that support it

These should provide some room for discussion about specifics how the group wants to abstract or break out functionality.

---

## Spatial backends based on GDAL

I imagine some users don't care so much what file format their target store contains, but likely will care more about the object types that are returned and the associated packages. The object type matters because of chosen dependencies and preferred workflows of the user. The file type may matter especially when it comes time to read targets back in, in part or in full, when they start taking up a lot of disk space, or some step in the process requires a specific format.

 - We may not want to require users to load both {sf} and {terra}, for example
 
   -  Package usage gated by `requireNamespace()` and having all of these types of packages that produce the user-facing object in _Suggests_ seems like a good strategy. The alternative would be to say, pick {terra} for use internally and then provide conversion methods for compatibility with other ({sf}/{stars}) objects as input/output.
   
[...]

PR #7, opened 2024-03-07
Rename tar_<filetype>() to tar_<pkg>_<data type>()

Aariq:
This PR renames `tar_geotiff()` to `tar_terra_rast()`.  It separates out the format into a new function, `format_terra_rast()` that returns the output of a `tar_format()` call. ~~I separated it out so it could accept a `filetype` argument where users can choose a GDAL driver (from a list of possibilities we have implemented and tested with `targets`).~~

I'm thinking maybe we'll end up having tar_ functions like `tar_<pkg name>_<data type/class>()` with some default storage formats rather than naming the functions after the file type.  What do you think?

Somewhat addresses #4


R/tar-terra.R
```diff
@@ -0,0 +1,54 @@
+#' @export
+tar_terra_rast <- function(name,
+                           command,
+                           pattern = NULL,
+                           packages = targets::tar_option_get("packages"),
+                           tidy_eval = targets::tar_option_get("tidy_eval"),
+                           library = targets::tar_option_get("library"),
+                           format = format_terra_rast(),
```
Aariq, line 8: This here is the main difference (sorry, looks like GitHub is registering this as an entirely  new file since I renamed it)

njtierney: Looks good to me!

njtierney:
I really like the idea of having  `tar_<pkg name>_<data type/class>()`. I think this will help encourage discoverability for users. Nice one!

R/formats.R
```diff
@@ -0,0 +1,30 @@
...
+    } else if (filetype == "netCDF") {
+        write_fun <- function(object, path) {
+            terra::writeRaster(
+                x = object,
+                filename = path,
+                overwrite = TRUE,
+                filetype = "netCDF"
+            )
+        }
+    }
```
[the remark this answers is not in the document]
Aariq: This is excellent, but I'm going to move it to an issue because I think there are API things to discuss before implementing this.

brownag (approved):
This looks good to me!

PR #9, opened 2024-03-08
Adds a WIP repostatus badge to readme and fixes R-CMD-check action

### LINKED ISSUE #8: use standard github action for check?
Looks like the current action is failing.  Would it be ok to replace it with the one that `usethis::use_github_action()` adds?

Aariq:
Closes #8


njtierney:
Thanks so much for this, @Aariq ! :)

Issue #16, opened 2024-03-13
Default behaviours for filetype and GDAL


njtierney:
As noted at https://github.com/njtierney/geotargets/blob/master/R/tar-terra-rast.R#L66-L73 by @brownag 

It could be a good idea to have some default values for things like raster filetypes ("GTiff", "netCDF", ?) and GDAL ("ENCODING=UTF-8").

These could be set with `options` or maybe there's a way {targets} like us to set target options - `tar_option_set()` doesn't look like it has slots for new arguments, so either we make our own `geotarget_option_set` function, or we can get the user to use `options` or set an envvar, or something. Note that https://design.tidyverse.org/def-user.html doesn't discuss this in great detail.

njtierney:
Fantastic, thank you so much! Looks great, just got some function name suggestions linked in the PR

That {covr} error was annoying, thanks for fixing that!

Aariq:
Closed by #19

PR #19, opened 2024-03-13
Add geotargets options functions, with defaults for raster GDAL driver and creation options

brownag:
Two new draft functions were added for #16 : `geotargets_option_get()` and `geotargets_option_set()`. 

I noticed that the code coverage was [failing](https://github.com/njtierney/geotargets/actions/runs/8258728257/job/22591465475). There seems to be an issue with modifying the function body for `create_format_terra_raster()` inside `covr::package_coverage()`. This PR fixes that.

It appears coverage issue can be resolved by allowing for the default options for GDAL driver (`filetype`) and creation options (`gdal`) to be used in the template function, rather than `NULL`. Though I am still curious what is happening inside the evaluation {covr} does that causes this to fail.


njtierney (changes requested):
Thanks so much for these changes, @brownag !  This is awesome.

Just a couple of suggestions regarding abstracting out some parts into functions - let me know if you would prefer not to do these?

I also think we will want to import `rlang` - it's already in `targets` so we aren't adding a dependency, and I think we could benefit from some rlang functions, such as `%||%`.

njtierney:
Also thank you so much for catching that {covr} error - I had just chalked it up to "CI be weird sometimes" and shelved it for later. This is a huge help!

R/geotargets-option.R
```diff
@@ -0,0 +1,50 @@
...
+               strsplit(Sys.getenv("GEOTARGETS_RASTER_GDAL_CREATION_OPTIONS",
+                                   unset = getOption(x, default = ifelse(is.null(value), "ENCODING=UTF-8", value))),
+                        ";")[[1]]
+           },
+           "geotargets.raster.gdal_driver_name" = {
+               Sys.getenv("GEOTARGETS_RASTER_GDAL_DRIVER_NAME",
+                          unset = getOption(x, default = ifelse(is.null(value), "GTiff", value))
+               )
+           })
+}
```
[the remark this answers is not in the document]
njtierney: Naming things is hard: some alternatives for these functions:

- get_geotargets_raster_gdal_creation_options
  - get_geotargets_gdal_options_raster_creation
  - get_geotargets_gdal_raster_options_creation

- get_geotargets_raster_gdal_driver_name
  - get_geotargets_gdal_name_raster_driver
  - get_geotargets_gdal_raster_name_driver

Thoughts welcome! I also appreciate that this is a small detail, so perhaps it isn't that important

R/geotargets-option.R
```diff
@@ -0,0 +1,50 @@
+#' Get or Set geotargets Options
+#'
+#' Get or set behavior for geospatial data target stores using geotargets-specific global options.
+#'
+#' @param x Character. Option name. See Details.
```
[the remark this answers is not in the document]
brownag: Sure

R/geotargets-option.R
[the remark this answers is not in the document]
brownag: Totally fine with an rlang import, but you'll have to forgive me when I don't think to use it while I am prototyping something. Good suggestion there and makes things clearer.

I also thought about putting `"_gdal_"` before the geospatial type in the option name, I would be fine with that. 

Breaking out these functions is a little more abstraction than I normally would do for something like this, but it may be good to have explicit setters for each option to facilitate setting them in a clear way. I ultimately don't have a problem with defining these specific functions other than we might wind up with a lot of them if we continue on this path

I imagine we could have analogous options for vector data, so the "raster" bit should come relatively early in the name so the user "branches early" for autocomplete etc. 

 I think I prefer either `get_geotargets_<raster/vector>_gdal_...()` or `get_geotargets_gdal_<raster/vector>_...()`

With "creation options" and "driver name", I am referring to GDAL constructs for these things (i.e. https://gdal.org/drivers/raster/gtiff.html#creation-options, https://gdal.org/drivers/raster/index.html). Trying to make it generic, not specifically based off of the argument names for terra functions. For example the sf analog of `terra::writeVector(..., filetype="foo", gdal="bar")` is `sf::st_write(..., driver="foo", layer_options="bar")`

R/tar-terra-rast.R
```diff
@@ -117,9 +117,9 @@ create_format_terra_raster <- function(filetype, gdal, ...) {
         terra::writeRaster(
             object,
             path,
-            filetype = NULL,
+            filetype = geotargets::geotargets_option_get("raster.gdal_driver_name"),
             overwrite = TRUE,
-            gdal = NULL
+            gdal = geotargets::geotargets_option_get("raster.gdal_creation_options")
```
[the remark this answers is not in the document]
brownag: The user specified `filetype` and `gdal` option specified by the user are injected after the function is defined, on L125 and L126, so that will overwrite any options if the user has set something for a specific call. With the caveat that that seems to not work in {covr} context--I will investigate this further as it may be a sign of a bigger problem with this approach.

https://github.com/njtierney/geotargets/blob/5b71a2afaa096dbcbb06481d6ab498734b9bba50/R/tar-terra-rast.R#L125-L126

The order of operations is a little confusing, so would be happy to revise if there is a clearer way. 
Currently in `tar_terra_rast()` if NULL user input we pull the option and check it against available drivers with `match.arg()`. If user input is specified, that is also checked against the drivers. The function template is defined to use defaults for drivers, which includes env vars and options, ultimately falling back when those are not defined to is loaded into `geotargets.env` on attaching the package

https://github.com/njtierney/geotargets/blob/5b71a2afaa096dbcbb06481d6ab498734b9bba50/R/AAAA.R#L7-L10

So, the precedence for source of these custom options goes:

 - User input to target function > System environment variable > global option > package environment

The failures in {covr} appear to be related to the `NULL` template values not being replaced by `body<-` when they should be. This seems like it could be some sort of issue with evaluation environments, or something funky {covr} does that delays evaluation of the `body<-` call.

> Also we don't need to use geotargets::

You are right. I intended to protect against the package namespace not being loaded in the context that the `tar_format(write=)` function gets called, but I confirm it at least doesn't break our current tests to remove this inside the `.write_terra_raster()` function

R/geotargets-option.R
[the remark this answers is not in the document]
brownag: I think I like the suggestion of putting "gdal" before "raster" or "vector"--it seems to be easier to read to me.

R/tar-terra-rast.R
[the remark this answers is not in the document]
brownag: Note that it seems that {covr} does not like us not having `geotargets::` inside the write function (https://github.com/njtierney/geotargets/pull/19/commits/f634c301432549d20f9da928ba9771839828fe90). But you were right the other cases I was using it were unneeded!

R/geotargets-option.R
```diff
@@ -0,0 +1,65 @@
...
+#' ## Available Options
+#'
+#'  - `"geotargets.gdal.raster.creation_options"` - set the GDAL creation options used when writing raster files to target store (default: `"ENCODING=UTF-8"`)
+#'
+#'  - `"geotargets.gdal.raster.driver_name"` - set the file type used for raster data in target store (default: `"GTiff"`)
+#'
+#'  Each option can be overridden with a system environment variable. Options include:
+#'
+#'   - `GEOTARGETS_GDAL_RASTER_CREATION_OPTIONS`
+#'   - `GEOTARGETS_GDAL_RASTER_DRIVER_NAME`
```
Aariq, line 18: Would it make sense to have these be a little shorter just for the sake of being concise?
```suggestion
#'   - `GEOTARGETS_GDAL_RASTER_OPTIONS`
#'   - `GEOTARGETS_GDAL_RASTER_DRIVER`
```

R/geotargets-option.R
```diff
@@ -0,0 +1,65 @@
...
+#'
+#' Get or set behavior for geospatial data target stores using geotargets-specific global options.
+#'
+#' @param option_name Character. Option name. See Details.
+#'
+#' @details
+#'
+#' ## Available Options
+#'
+#'  - `"geotargets.gdal.raster.creation_options"` - set the GDAL creation options used when writing raster files to target store (default: `"ENCODING=UTF-8"`)
```
Aariq, line 11: Can one supply multiple options?  If so, how should they be delimited?

R/geotargets-option.R
```diff
@@ -0,0 +1,65 @@
...
+#'   - `GEOTARGETS_GDAL_RASTER_CREATION_OPTIONS`
+#'   - `GEOTARGETS_GDAL_RASTER_DRIVER_NAME`
+#'
+#' @rdname geotargets-options
+#' @export
+geotargets_option_get <- function(option_name) {
+    if (!startsWith(option_name, "geotargets.")) {
+        option_name <- paste0("geotargets.", option_name)
+    }
+
```
Aariq, line 26:
```suggestion
targets:::tar_assert_flag(option_name, choices = names(geotargets.env))
```
Or some other method of erroring when an invalid option name is provided

Aariq (changes requested):
A few ideas/suggestions and a request to get an error if you ask to set/get an invalid option name.

R/geotargets-option.R
```diff
@@ -0,0 +1,50 @@
...
+           })
+}
+
+#' @param value Value to assign to option `x`.
+#' @rdname geotargets-options
+#' @export
+geotargets_option_set <- function(x, value) {
+    if (!startsWith(x, "geotargets.")) {
+        x <- paste0("geotargets.", x)
+    }
```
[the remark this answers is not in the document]
njtierney: It's always a bit of a balance trying to create magic for the users - @milesmcbain has a great idea that magic = power/pain. I think we are removing pain from the user here by not needing to type out long option names, but we can always revisit it later if we need to.

Agree that global option/system envvars probably need to have specific geotargets prefix

R/tar-terra-rast.R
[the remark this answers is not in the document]
njtierney: Awesome thanks for explaining! 

> User input to target function > System environment variable > global option > package environment

Makes sense to me, this sort of stuff is a bit curly, so I appreciate you unpacking it, and I think it's yet another reason to have this package to help remove some of this heavy lifting.

R/geotargets-option.R
njtierney: I think OPTIONS is open ended, and could refer to other parts of this, e.g., "open-options": https://gdal.org/drivers/raster/gtiff.html#open-options

While CREATION_OPTIONS is longer, and I want to reduce typing for the user, I think it might be necessary?

I think RASTER_DRIVER makes sense

R/geotargets-option.R
njtierney: Good idea!

```r
rlang::arg_match0(option_name, names(geotargets.env)
```

should work, and means we won't need to do a (CRAN illegal) `targets:::`

R/geotargets-option.R
njtierney: I think that maybe we can open this discussion as a separate issue? I'm keen to implement this PR

njtierney: Opened #25

R/geotargets-option.R
njtierney: Great point, but I think this might be better as a separate issue: https://github.com/njtierney/geotargets/issues/26

R/geotargets-option.R
njtierney: broke this out into a separate issue: https://github.com/njtierney/geotargets/issues/27

njtierney (approved):

Aariq (approved):

Issue #10, opened 2024-03-11
Move all supported r-spatial packages to Suggests


Aariq:
In #4 @brownag mentioned moving `terra` and other r-spatial packages from Imports to Suggests, and I think this is a good call.  We shouldn't require users to have all supported r-spatial packages installed just to use `targets` with one of them.   `rlang::check_installed()` prompts users to install a missing package and installs it for them, but since this code will mostly be run in a `callr` environment, it might be more appropriate to just use `rlang::is_installed()`.  We could even supply a custom tip like 
>! package `terra` is required. 
>Did you forget to add "terra" to `tar_option_set(packages = `?

PR #21, opened 2024-03-13
Add filetype argument to `tar_terra_vect()`

Aariq:
I tried to replicate additions to `tar_terra_rast()` in `tar_terra_vect()`.  I think ESRI Shapfiles need special handling, and potentially other formats won't work.  I made the default format GeoJSON simply because it is an included driver and produces a single file.


Aariq:
Not quite working yet.  Somewhere I've not defined the object `gdal`. Can investigate tomorrow

njtierney:
I perhaps wonder if once #19 is merged and this is rebased/merged off of that if it will help resolve some of those changes?

R/tar-terra-vect.R
```diff
@@ -67,89 +71,75 @@
...
+    if (is.null(gdal)) {
+        gdal <- "ENCODING=UTF-8"
+    }
+
+    if (filetype == "ESRI Shapefile") {
+        #special handling of ESRI shapefiles because the output is a dir of multiple files.
+        format <- create_format_terra_vect_shz(options = gdal, ...)
+    } else {
+        format <- create_format_terra_vect(filetype, options = gdal, ...)
+    }
```
njtierney, line 122: I would say put this into an ifelse, like so:

```suggestion
format <- ifelse(
  test = filetype == "ESRI Shapefile",
  #special handling of ESRI shapefiles because the output is a dir of multiple files.
  yes = create_format_terra_vect_shz(options = gdal, ...),
  no =  create_format_terra_vect(filetype, options = gdal, ...)
)
```

If we want to use more than 2, then I think `switch()` is worthwhile

njtierney (approved):
Looks good to me - one small suggestions on if vs ifelse, what do you think?

R/tar-terra-vect.R
Aariq: I haven't done a thorough exploration of what other drivers might need special handling like this, so this is fine with me

njtierney:
It passes checks! Did you want to merge now?

Aariq:
If you'd rather merge #19 first, I can update this PR to use default options for filetype.  Or we can merge them both and do that in a separate PR.  Doesn't matter to me.

njtierney:
I think I'd prefer to merge #19 first actually - there are a couple of changes that you requested, I'll comment on them in the PR - but maybe these could be addressed in separate issues.

R/tar-terra-rast.R
```diff
@@ -1,9 +1,9 @@
 #' Create a terra _SpatRaster_ Target
 #'
-#' Creates a target for a terra _SpatRaster_ object.
+#' Provides a target forma for [terra::SpatRaster-class] objects.
```
brownag, line 3:
```suggestion
#' Provides a target format for [terra::SpatRaster-class] objects.
```

brownag (approved):
Overall looks good to me, just spotted one tiny typo. Thanks for adding the links too!

We will need to create options for vector data to fix the {covr} failure, and should hash out the naming in #25

I spent a little bit looking into the mechanics of {covr} last night... for now I do not have an alternate solution to what was done in #19, so will want to use defaults via the `geotargets_option_get()` approach for vector data too

R/tar-terra-vect.R
```diff
@@ -29,68 +36,144 @@
...
-    priority = priority,
-    resources = resources,
-    storage = storage,
-    retrieval = retrieval,
-    cue = cue
-  )
+                           command,
+                           pattern = NULL,
+                           filetype = NULL,
+                           gdal = NULL,
```
Aariq, line 42: Is there any reason not to do things like this?
```suggestion
                           filetype = geotargets_option_get("gdal.raster.driver"),
                           gdal = geotargets_option_get("gdal.raster.creation_options"),
```
I think that would simplify the logic of the unexported functions and make it more obvious how users can change the defaults.

R/tar-terra-vect.R
```diff
@@ -29,68 +36,144 @@
...
+
+#' @param filetype File format expressed as GDAL driver names passed to `terra::writeVector()`
+#' @param options GDAL driver specific datasource creation options passed to `terra::writeVector()`
+#' @param ... Additional arguments not yet used
+#' @noRd
+create_format_terra_vect <- function(filetype, options, ...) {
+
+    if (!requireNamespace("terra")) {
+        stop("package 'terra' is required", call. = FALSE)
+    }
```
Aariq, line 119: I feel like this should be in the exported `tar_*()` function only.

R/tar-terra-vect.R
Aariq: Opened an issue for this (#34)

R/tar-terra-vect.R
Aariq: I'll do this in a separate PR that will close #10

Issue #25, opened 2024-03-15
Naming geotargets gdal raster options


njtierney:
Breaking this out of #19 

Naming things is always a bit tricky, but following on from @Aariq 's suggestion on renaming some of the env vars, current thoughts are:

`GEOTARGETS_GDAL_RASTER_DRIVER_NAME` -- > `GEOTARGETS_GDAL_RASTER_DRIVER`

It's a small change but I think it's worthwhile to reduce typing where practical. On the [GDAL website they are called "Raster drivers"](https://gdal.org/drivers/raster/index.html) too, so there's that?

brownag:
Totally agree with dropping "_NAME" from the option name, that is a great suggestion

I also agree with your point RE: "CREATION_OPTIONS" (https://github.com/njtierney/geotargets/pull/19#discussion_r1524052254).

At a minimum we need to distinguish between options used to write/create a datasource vs. those used to read them.

njtierney:
closed in #30

Issue #26, opened 2024-03-15
Linking to documentation for specifying multiple options in `geotargets.gdal.raster.creation_options`


njtierney:
As asked by @Aariq in #19 

> Can one supply multiple options? If so, how should they be delimited?

In reference to:

```r
#' ## Available Options
#'
#'  - `"geotargets.gdal.raster.creation_options"` - set the GDAL creation options used when writing raster files to target store (default: `"ENCODING=UTF-8"`)
```

It might would be useful to discuss how to specify multiple options, and where users can go to find more options to set. E.g., for setting multiple options we could reference this:

```r
geotargets::geotargets_option_set("raster_gdal_creation_options", c("COMPRESS=DEFLATE", "TFW=YES"))
```

And if we want users to have the capacity to provide multiple options, maybe we point them to https://gdal.org/drivers/raster/gtiff.html#creation-options to list creation options? Is that the right source?

brownag:
Thanks @Aariq for pointing this out, I had meant to add a blurb to document this then forgot!

For options we can specify a character vector with length >= 1 where each element has form `"OPTION=VALUE"`

For the system environment variables we don't have the ability to have length > 1, so I hardcoded the use of semicolon as a delimiter, so when `Sys.getenv()` is called the result is `strsplit()` using `";"`. Open to alternatives to the latter, but probably just needs to be documented for now. It may be that we don't really _need_ system environment variables at all, but it may be useful for some folks to be able to do it that way.

RE: the source to reference. This would vary depending on the driver being used, but could certainly give GeoTIFF driver as an example. Other raster drivers may differ in the creation options they support, e.g. https://gdal.org/drivers/raster/netcdf.html#creation-options

njtierney:
closed in #30

Issue #27, opened 2024-03-15
Check if invalid option names are provided in `geotargets_option_get()`


njtierney:
As stated by @Aariq in #19 

> ...[give] some other method of erroring when an invalid option name is provided

Referring to after this code chunk:

```r
geotargets_option_get <- function(option_name) {
    if (!startsWith(option_name, "geotargets.")) {
        option_name <- paste0("geotargets.", option_name)
    }

# ...
```

We could do something like:

```r
rlang::arg_match0(option_name, names(geotargets.env)
```

Aariq:
I think this is fixed now

PR #30, opened 2024-03-15
geotargets options / defaults update

brownag:
Some updates to options following #19 for #26 

 - Adds vector data analogs for options introduced in #19, can be used for #21
 - Updates "driver name" options to remove _NAME/_name suffix for #25
 - Other cleanup: fix some option names not updated in #19, and only specify default arguments once (in `geotargets_option_get()`

### DIFF
--- a/R/geotargets-option.R
+++ b/R/geotargets-option.R
@@ -8,14 +8,22 @@
 #'
 #' ## Available Options
 #'
-#'  - `"geotargets.gdal.raster.creation_options"` - set the GDAL creation options used when writing raster files to target store (default: `"ENCODING=UTF-8"`)
+#'  - `"geotargets.gdal.raster.driver"` - character. Length 1. Set the driver used for raster data in target store (default: `"GTiff"`). Options for driver names can be found here: <https://gdal.org/drivers/raster/index.html>
 #'
-#'  - `"geotargets.gdal.raster.driver"` - set the file type used for raster data in target store (default: `"GTiff"`)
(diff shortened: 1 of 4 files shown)


R/geotargets-option.R
```diff
@@ -39,27 +45,51 @@ geotargets_option_get <- function(option_name) {
...
         )
     }
 
+    get_geotargets_gdal_vector_creation_options <- function(option_name, option_value) {
+        gdal_creation_options <- Sys.getenv(
+            x = "GEOTARGETS_GDAL_VECTOR_CREATION_OPTIONS",
+            unset = get_option(option_name, option_value, "ENCODING=UTF-8")
+        )
+        the_option <- strsplit(gdal_creation_options, ";")[[1]]
+        the_option
```
njtierney, line 61:
```suggestion
        some_options <- strsplit(gdal_creation_options, ";")[[1]]
        some_options
```

njtierney (approved):
Looks good to me!

njtierney:
Really nice work! I believe this resolves #25 #26 ?

It looks like it handles all we need for those issues to me. We can come back to #27 in a separate PR I think. Are you happy for me to merge?

brownag:
> Really nice work! I believe this resolves #25 #26 ?
> 
> It looks like it handles all we need for those issues to me. We can come back to #27 in a separate PR I think. Are you happy for me to merge?

Thanks. Last thing was I realized that ENCODING as a default creation option for GTiff doesn't make sense... that is the default for terra::writeVector() but few or no raster drivers support it anyway.

njtierney:
Yup! Awesome 🚀

Issue #14, opened 2024-03-12
Implement tar_stars_* family of targets


njtierney:
Related to #4 

We need to list the datatypes from {stars} that we want to support

brownag:
In #33 I started by implementing a function `tar_stars()` based on `stars::write_stars()` to support targets based on stars objects. I also added an extension to the `tar_format()` read function to support stars_proxy objects after @Aariq's suggestion

It is possible to store both raster and vector data in stars objects. For multidimensional arrays, which can be derived from vector data, a driver needs to be chosen that will support multidimensional array output (e.g. netCDF, Zarr). 

`stars::write_mdim()` can be used to write using the GDAL Multidimensional Raster Data Model. https://gdal.org/user/multidim_raster_data_model.html. 

I implemented a sketch of this in https://github.com/njtierney/geotargets/pull/33/commits/f74fe91d48f6c6fdab5ffb13e6895bfb555c9a6c which implements the `mdim` argument to toggle the multidimensional API in `tar_stars()` and `tar_stars_proxy()`

We could add options specifically for "mdim"-related processing such as `"gdal.mdim.driver"`, since the user may wish to be able to handle their multidimensional datasets differently than other 2D raster data or vector data.

Aariq:
Would this be closed by #33 or are there other `tar_stars*()` functions we're imagining?

Aariq:
I'm going to close this.  Feel free to open a follow-up issue if there are other `tar_stars_*` functions that are needed.

PR #33, opened 2024-03-15
Add `tar_stars()`

brownag:
Adds format methods `tar_stars()`, `tar_stars_proxy()` for #14 

 - Adds {sf}, {stars} and {ncmeta} to suggests
 - When `mdim=TRUE` `tar_stars()` and `tar_stars_proxy()` can use the GDAL Multidimensional Raster Data Model as a backend via `stars::read/write_mdim()`instead of `stars::read/write_stars()`
 - When `ncdf=TRUE` "read" methods can be toggled to use `stars::read_ncdf()` (instead of `stars::read_stars()` or `stars::read_mdim()`) which requires {ncmeta} 

 - To be less repetitive I think`tar_stars_proxy()` could call `tar_stars(proxy=TRUE)` but I still need to sort out how arguments would need to be deparsed to do this.


R/tar-stars.R
```diff
@@ -0,0 +1,136 @@
...
+}
+#' @param driver character. File format expressed as GDAL driver names passed to [stars::write_stars()]. See [sf::st_drivers()].
+#' @param options character. GDAL driver specific datasource creation options passed to [stars::write_stars()]
+#' @param ... Additional arguments not yet used
+#' @noRd
+create_format_stars <- function(driver, options, ...) {
+
+    if (!requireNamespace("sf")) {
+        stop("package 'sf' is required", call. = FALSE)
+    }
```
Aariq, line 98: `sf` is in Depends for `stars`, so this is probably not necessary

R/tar-stars.R
```diff
@@ -0,0 +1,136 @@
...
+
+    if (!requireNamespace("stars")) {
+        stop("package 'stars' is required", call. = FALSE)
+    }
```
Aariq, line 102: Maybe move this to the very top of `tar_stars()`

R/tar-stars.R
```diff
@@ -0,0 +1,136 @@
...
+    # NOTE: Option getting functions are set in the .write_terra_raster function template
+    #       to resolve issue with body<- not working in some evaluation contexts ({covr}).
+    # TODO: It should be fine to have driver and options as NULL
+    .write_stars <- function(object, path) {
+        stars::write_stars(
+            object,
+            path,
+            driver = geotargets::geotargets_option_get("gdal.raster.driver"),
+            overwrite = TRUE,
+            options = geotargets::geotargets_option_get("gdal.raster.creation_options")
```
Aariq, line 122: Can't `stars` objects contain vectors?  How does this work with vector data cubes?

tests/testthat/test-tar-stars.R
```diff
@@ -0,0 +1,21 @@
+# test_that() #Included to make RStudio recognize this file as a test
+
+skip_if_not_installed("sf")
```
Aariq, line 3: I think this is unecessary, again because `sf` is a hard dependency of `stars`

Aariq (commented):
It might be good to clarify that this currently works for stars raster cubes, but might not work for vector cubes or stars.proxy objects. We might want to explore these other formats a bit and depending on the findings of that exploration, perhaps call this `tar_stars_raster` even though both raster cubes and vector cubes have the same class (`stars`).

R/tar-stars.R
brownag: You are right, it was a reflex for me b/c I was using `sf::st_drivers()`, thanks for pointing this out, will fix

R/tar-stars.R
brownag: will do to mirror your other changes/suggestions!

R/tar-stars.R
```diff
@@ -0,0 +1,136 @@
+#' Create a stars _stars_ Target
+#'
+#' Provides a target format for stars objects.
+#'
+#' @param driver character. File format expressed as GDAL driver names passed to [stars::write_stars()]. See [sf::st_drivers()].
+#' @param options character. GDAL driver specific datasource creation options passed to [stars::write_stars()]
+#' @param ... Additional arguments not yet used
+#'
+#' @inheritParams targets::tar_target
+#' @importFrom rlang %||% arg_match0 is_installed
```
njtierney, line 10: We don't need to use `#' @importFrom rlang <fun>` each time we want to use those functions - I suggest that we move these to https://github.com/njtierney/geotargets/blob/master/R/geotargets-package.R#L4-L5

R/tar-stars.R
njtierney, line 102:
```suggestion
check_installed("stars")
```

njtierney (changes requested):
some suggestions on pkg documentation and using `rlang::check_installed()`

R/tar-stars.R
brownag: done in https://github.com/njtierney/geotargets/pull/33/commits/0e20c172f574efed855625ec06219c40eade9917

R/tar-stars.R
brownag: done in https://github.com/njtierney/geotargets/pull/33/commits/0e20c172f574efed855625ec06219c40eade9917 and https://github.com/njtierney/geotargets/pull/33/commits/78bdee5347f99818837669bac4aa75e69911f6f4

tests/testthat/test-tar-stars.R
brownag: done in https://github.com/njtierney/geotargets/pull/33/commits/0e20c172f574efed855625ec06219c40eade9917

R/tar-stars.R
brownag: done in https://github.com/njtierney/geotargets/pull/33/commits/0e20c172f574efed855625ec06219c40eade9917

R/tar-stars.R
brownag: done in https://github.com/njtierney/geotargets/pull/33/commits/0e20c172f574efed855625ec06219c40eade9917

brownag:
> It might be good to clarify that this currently works for stars raster cubes, but might not work for vector cubes or stars.proxy objects. 

Good point about vector cubes, we should work up an example for that in a new issue. We can hash it out before merging this PR if need be.

In https://github.com/njtierney/geotargets/pull/33/commits/0e20c172f574efed855625ec06219c40eade9917 I added support for `stars_proxy` objects. `tar_stars()` now takes a `proxy` argument (default: FALSE) and there is a `tar_stars_proxy()` function for shorthand.  I _think_ `tar_stars_proxy()` could call `tar_stars()` but would need to think more about the deparsing/evaluation of the arguments to make it work.

> We might want to explore these other formats a bit and depending on the findings of that exploration, perhaps call this `tar_stars_raster` even though both raster cubes and vector cubes have the same class (`stars`).

I think we should open an issue for this and look into some vector data cube examples--I am not a heavy user of stars, and can't say I a ton of experience with that specific application--though I want to investigate now. 

As I understand it creating vector data cubes requires reading the vector data with `sf::st_read()`/`sf::read_sf()` and then converting to stars with `stars::st_as_stars()`. So, we definitely might want to provide a target that eases this process, but right now if the user has a stars or stars_proxy object already made, I _think_ what is implemented in this PR "works"--open to a counterexample!

R/tar-stars.R
brownag: Yes, good point, `stars` objects can contain vector geometry and an arbitrary number of dimensions. I said "new" issue but we can discuss this approach more in #14, I think

R/tar-stars.R
```diff
@@ -0,0 +1,291 @@
+#' Create a stars _stars_ Target
+#'
+#' Provides a target format for stars objects.
+#'
+#' @param proxy logical. Passed to [stars::read_stars()]. If `TRUE` the target an object of class `stars_proxy`. Otherwise, the object is class `stars`.
+#' @param mdim logical. Use the [Multidimensional Raster Data Model](https://gdal.org/user/multidim_raster_data_model.html) via [stars::write_mdim()]? Default: `FALSE`. Only supported for some drivers, e.g. `"HDF5"`, `"netCDF"`, `"Zarr"`.
+#' @param driver character. File format expressed as GDAL driver names passed to [stars::write_stars()] or [stars::write_mdim()] See [sf::st_drivers()].
+#' @param options character. GDAL driver specific datasource creation options passed to [stars::write_stars()]
```
Aariq, line 8: Just noticing that sometimes this arg is called `options` and sometimes it is called `gdal`.  We should decide whether we want to be consistent throughout geotargets or if we want to be consistent with whatever function the arg is being passed to (or split the difference and call it `gdal_options`)

brownag: @Aariq 

> Just noticing that sometimes this arg is called `options` and sometimes it is called `gdal`. We should decide whether we want to be consistent throughout geotargets or if we want to be consistent with whatever function the arg is being passed to (or split the difference and call it `gdal_options`)

This is a good idea. I was sortof leaning towards emulating whatever the "write" function uses for each particular spatial backend being used... But that won't cover everything. I definitely see the value of choosing something consistent across geotargets.

brownag: Resolving this, lets discuss in #44

Aariq:
If you'd like, I can attempt to bring this up to speed with the master branch.  That'll make it easier to review and discuss.

brownag:
> If you'd like, I can attempt to bring this up to speed with the master branch. That'll make it easier to review and discuss.

Thanks for the offer, I've been trying to keep up with changes... and wanted to make sure I went over everything so got the new master branch merged in here.

I am going to re-request review on this now. I think all prior comments/changes have been resolved and/or referred to in an open issue.

R/tar-stars.R
```diff
@@ -0,0 +1,220 @@
+#' Create a stars _stars_ Target
+#'
+#' Provides a target format for stars objects.
+#'
+#' @param driver character. File format expressed as GDAL driver names passed to [stars::write_stars()]. See [sf::st_drivers()].
+#' @param options character. GDAL driver specific datasource creation options passed to [stars::write_stars()]
+#' @param proxy logical. Passed to [stars::read_stars()]. If `TRUE` the target an object of class `stars_proxy`. Otherwise, the object is class `stars`.
```
Aariq, line 7:
```suggestion
#' @param proxy logical. Passed to [stars::read_stars()]. If `TRUE` the target will be read as an object of class `stars_proxy`. Otherwise, the object is class `stars`.
```

Seems like there is a missing word here maybe?

R/tar-stars.R
```diff
@@ -0,0 +1,220 @@
...
+        FUN(
+            object,
+            path,
+            driver = driver,
+            overwrite = TRUE,
+            options = options
+        )
+    }, list(FUN = str2lang(WRITE_FUN),
+            driver = driver,
+            options = options)))
```
Aariq, line 211: Rather than `eval(substitute(...` we now have the option to pass env vars to the custom format using `targets::tar_resources_custom_format()`.  See the updates to `tar_terra_rast()`, e.g. [here](https://github.com/njtierney/geotargets/blob/e1b039f372f5f82c9cd08ac27c319c08955f0c75/R/tar-terra-rast.R#L85-L91) and [here](https://github.com/njtierney/geotargets/blob/e1b039f372f5f82c9cd08ac27c319c08955f0c75/R/tar-terra-rast.R#L104-L108).  I think we should stick to this new pattern as it's a bit easier to comprehend, IMO.

Aariq (changes requested):
Could you re-run devtools::document() to update tar_stars.Rd

R/tar-stars.R
brownag: > Rather than `eval(substitute(...` we now have the option to pass env vars to the custom format using `targets::tar_resources_custom_format()`. 

I have this mostly implemented, it does make the code look cleaner. 

I'm running into some issues with arguments not being properly passed when using `tar_resources_custom_format()` though. I tinkered with it a bit last night and will take a fresh look tonight. If I can't figure it out I will push the changes anyway to see if you can spot what I am doing wrong.

R/geotargets-package.R
```diff
@@ -7,6 +7,7 @@ NULL
 
 # clear CRAN checks spotting floating global variables
 #' @importFrom utils globalVariables
+#' @importFrom rlang %||% arg_match0 is_installed check_installed
```
njtierney, line 10: importing `arg_match0`, `is_installed` and `check_installed` with `@importFrom` is only needed if we want to use those functions without namespacing them. My preference is that we don't @importFrom these functions as they aren't used very often.

It is necessary to importFrom `%||%` at the moment, although I note that in R 4.4.0 `%||%` is implemented natively, which is very cool.

tests/testthat/test-tar-stars.R
njtierney, line 1: Could you add one of these uses of tar_tars into the README.Rmd?

njtierney (changes requested):
Pending the changes from @Aariq and a minor change that I have about using `@importFrom` and a request to add an example of stars into the README, I think this is looking good :)

tests/testthat/test-tar-stars.R
brownag: I have now updated NAMESPACE, added a tar_stars() example and did a several other general updates to README

njtierney (approved):

njtierney:
Fantastic, thanks @brownag ! :)

R/tar-stars.R
brownag: It appears the source of my problem is that environment variables that are created via `tar_resources_custom_format()` are not available / not working properly in the `tar_format()` `read` method. 

I managed to localize the problems I was having to that, so for time being continuing to use `eval(substitute(...))` for read method dynamics. Using the env vars for write method, though. 

I have not yet had a chance to look into whether this is "expected" or if it is an issue in targets.

All tests are passing with expected values for the current code. If code to access env vars in read method is uncommented, the `tar_stars_proxy()` tests fail because default `proxy=FALSE` is used because the env var GEOTARGETS_PROXY is apparently unset.

R/geotargets-package.R
brownag: I have removed all three from importFrom as requested.. Indeed, only using `is_installed()` for `@examplesIf` block to check for stars, which may not really be necessary but I think we don't want to assume suggested packages are installed, in general. `check_installed()` is not used but I initially thought we would use it, thanks for pointing that out. `arg_match0` is used in a couple places now for terra and stars drivers checks.

For imports I generally prefer to handle functions explicitly, as well as use double colon to access them; the latter because it is easier for humans to see what is coming from where, and the former to explicitly declare what is used by the package in the NAMESPACE--no matter how commonly it is used.

njtierney: Thank you! :)

> For imports I generally prefer to handle functions explicitly, as well as use double colon to access them; the latter because it is easier for humans to see what is coming from where, and the former to explicitly declare what is used by the package in the NAMESPACE--no matter how commonly it is used.

I hear you! I prefer to use `pkg::fun()`, with the only exception being for imported infix operators, e.g., `%||%`. Another compelling case I've seen is for packages that make heavy use of ggplot2 functions or rlang metaprogramming, then you can use `@import ggplot2` etc.

 > explicitly declare what is used by the package in the NAMESPACE--no matter how commonly it is used.

I think this is a good idea, and it can be helpful to identify which functions are actually used from a package. It does have the potential to get a bit gnarly if you are making heavy use of importing from various packages though. I haven't seen much more development of the [itdepends](https://github.com/r-lib/itdepends) pkg, but I think tools like that would be really useful. The reason I don't use @importFrom very often is because of reasons stated in the [rpkgs book section on dependencies](https://r-pkgs.org/dependencies-in-practice.html#sec-dependencies-in-imports-r-code):

> @importFrom aaapkg aaa_fun. This does suppress the NOTE, but it also does more. This causes aaapkg to be loaded whenever your package is loaded. In contrast, if you use the approach we recommend, the aaapkg will only be loaded if your user does something that actually requires it. This rarely matters in practice, but it’s always nice to minimize or delay the loading of additional packages.

Anyway, that is a bit of an aside! I want to say thanks again for all your continued help with this project. I'm now curious if the `itdepends` project can be revived 🤔

R/tar-stars.R
```diff
@@ -0,0 +1,283 @@
...
+            custom_format = targets::tar_resources_custom_format(
+                #these envvars are used in read and write functions of format
+                envvars = c("GEOTARGETS_GDAL_RASTER_DRIVER" = driver,
+                            "GEOTARGETS_GDAL_RASTER_CREATION_OPTIONS" =
+                                paste0(options, collapse = ";"),
+                            "GEOTARGETS_GDAL_RASTER_MDIM" = mdim,
+                            "GEOTARGETS_PROXY" = proxy,
+                            "GEOTARGETS_USE_NCMETA" = ncdf)
+            )
+        ),
```
Aariq, line 278: Use user-specified `resources` as in #67 
```suggestion
        resources = utils::modifyList(targets::tar_resources(
            custom_format = targets::tar_resources_custom_format(
                #these envvars are used in read and write functions of format
                envvars = c("GEOTARGETS_GDAL_RASTER_DRIVER" = driver,
                            "GEOTARGETS_GDAL_RASTER_CREATION_OPTIONS" =
                                paste0(options, collapse = ";"),
                            "GEOTARGETS_GDAL_RASTER_MDIM" = mdim,
                            "GEOTARGETS_PROXY" = proxy,
                            "GEOTARGETS_USE_NCMETA" = ncdf)
            )
        ), resources)
```

Aariq (changes requested):
Looking good.  Only other thing is can you update NEWS.md now that we have it?

brownag:
> Looking good. Only other thing is can you update NEWS.md now that we have it?

Thanks, done!

R/tar-stars.R
Aariq: I see that you've commented this well, so I'm going to go ahead and merge this PR.  I think once we have all had some more practice with metaprogramming with targets, we'll go through and harmonize things across functions a bit.

Issue #38, opened 2024-03-16
Apparent bugs / problems with function body modification with covr+testthat+tar_test()


brownag:
Replacement of a function body does not seem to work properly within the evaluation environment created by combination of {covr}, targets tests with `tar_test()`, and {testthat}. We dynamically modify our internal functions using `body<-` to customize the functions passed to `tar_format()`.

Interactive runs, local runs, runs of {testthat} locally and remotely will all work fine, but when running tests via {covr} the problem arises.

I did some digging the other day into this issue, and there is a great vignette on how {covr} works behind the scenes: https://cran.r-project.org/web/packages/covr/vignettes/how_it_works.html

From the vignette:
> The core function in covr is `trace_calls()`. This function was adapted from ideas in [Advanced R - Walking the Abstract Syntax Tree with recursive functions](http://adv-r.had.co.nz/Expressions.html#ast-funs). This recursive function modifies each of the leaves (atomic or name objects) of an R expression by applying a given function to them. If the expression is not a leaf the walker function calls itself recursively on elements of the expression instead.

My suspicion is whatever {covr} does to modify the code during its processing causes a situation where no modification to the `tar_format()` read/write function occurs when we call `body<-`.

To solve this, I think we need to isolate the bug in a reprex and report it to {covr} maintainers. Probably this is not intended behavior, since other evaluation contexts have no apparent issues. In the meantime we could possibly find a way to "protect" or exclude the specific functions from being modified by {covr}.

I have tried several different alternatives to current approach--including converting function body to string and modifying the string, replacing the whole function; using an intermediate object; storing the function in a different environment and modifying it there--and no method of replacement with `body<-` seems to work.

These tests indicate to me it is not something simple like that the {covr} modified function body has more than 2 elements, or that the names of the arguments are missing, or that it is evaluating in the wrong environment. It seems to mealmost like there is a copy of the function made, and it corresponds to the initial function definition, and the body calls do not get applied to the right version of the function, which is what is ultimately passed to `tar_format()`

@Aariq had some good thoughts on covr related issues worth following up on:

> Would it be helpful to maybe look at how `targets`, `tarchetypes`, etc. deals with this?  There might be some hints in their github actions or in commit history.  I'm assuming the coverage issues have something to do with `callr` R sessions in tests, but I don't actually know.  Not sure if this is helpful, but I've noticed that the tests use the *installed* version of `geotargets` because of the namepacing with `geotargets::` rather than using the current state of the code.

[...]

Issue #11, opened 2024-03-12
Implement "filetype" argument


Aariq:
If `tar_*` functions are specific to packages and data types, then adding a `filetype` argument somewhere would be a way for users to override defaults for what kind of file targets are stored as (e.g. GeoTIFF vs netCDF).  I could imagine `filetype` being an argument to `tar_terra_rast()` or an argument to a function supplied to the `format` argument of `tar_terra_rast()`

For example:
Option 1
``` r
tar_terra_rast <-
  function(name, command, pattern = NULL, filetype = c("GeoTIFF", "netCDF"), ...)
```

Option 2

```r
tar_terra_rast <-
  function(name, command, pattern = NULL, format = format_terra_rast(filetype = c("GeoTIFF", "netCDF")), ...)
```
where `format_terra_rast()` returns the result of a call to `tar_format()`

The first option is probably preferable, unless there are other customizations that users might need to do to the format

Aariq:
Related comment from PR #7:

I had a bit of a play and I couldn't quite come up with one, I thought you could pass the argument in and R's lexical scoping would handle it, but not quite.

I think this might be another way to implement this - it is initially more code, but the core of the function I think becomes easier to extend. Thank you to @maelle for showing me https://rlang.r-lib.org/reference/arg_match.html

``` r
write_raster_gtiff <- function(object, path) {
    function(object, path) {
        terra::writeRaster(
            x = object,
            filename = path,
            overwrite = TRUE,
            filetype = "GTiff"
        )
    }
}

write_raster_netcdf <- function(object, path) {
    function(object, path) {
        terra::writeRaster(
            x = object,
            filename = path,
            overwrite = TRUE,
            filetype = "netCDF"
        )
    }
}

create_write_fun <- function(filetype = c("GTiff", "netCDF")) {
    rlang::arg_match(filetype)
    switch(filetype,
           "GTiff" = write_raster_netcdf(filetype),
           "netCDF" = write_raster_gtiff(filetype)
    )
}

create_write_fun("GTiff")
#> function(object, path) {
#>         terra::writeRaster(
#>             x = object,
#>             filename = path,
#>             overwrite = TRUE,
#>             filetype = "netCDF"
#>         )
#>     }
#> <environment: 0x14f561368>
create_write_fun("netCDF")
#> function(object, path) {
#>         terra::writeRaster(
#>             x = object,
#>             filename = path,
#>             overwrite = TRUE,
#>             filetype = "GTiff"
#>         )
[...]

PR #39, opened 2024-03-16
create_format_terra_vect: temp fix for issue with replacing function body and {covr}

brownag:
This is a workaround for replacement of a function body not working properly within the evaluation environment created by covr, targets tests, and testthat.

More commentary in new issue #38

### DIFF
--- a/R/tar-terra-vect.R
+++ b/R/tar-terra-vect.R
@@ -131,9 +131,9 @@ create_format_terra_vect <- function(filetype, options, ...) {
         terra::writeVector(
             object,
             path,
-            filetype = NULL,
+            filetype = geotargets::geotargets_option_get("gdal.vector.driver"),
             overwrite = TRUE,
-            options = NULL
+            options = geotargets::geotargets_option_get("gdal.vector.creation_options")
         )
     }
     body(.write_terra_vector)[[2]][["filetype"]] <- filetype


R/tar-terra-vect.R
```diff
@@ -131,9 +131,9 @@ create_format_terra_vect <- function(filetype, options, ...) {
...
             object,
             path,
-            filetype = NULL,
+            filetype = geotargets::geotargets_option_get("gdal.vector.driver"),
             overwrite = TRUE,
-            options = NULL
+            options = geotargets::geotargets_option_get("gdal.vector.creation_options")
         )
     }
     body(.write_terra_vector)[[2]][["filetype"]] <- filetype
```
njtierney, line 139: I wonder if we even actually need to modify the function body directly? I'm fairly sure those parts will be evaluated appropriately?

E.g.,

``` r
example_funfun <- function(sep_type){
  .write_fun <- function(object, file){
    # loudly declare the quote type when writing
    readr::write_lines(x = object,
                       file = file,
                       sep = sep_type)
    cat("Hi! The separation marker is: '", sep_type,"'", sep = "")
    cat("\n")
    cat("We wrote the file out to:\n", file, sep = "")
  }
  
  return(
    list(
      read = readr::read_csv,
      write = .write_fun
    )
  )
  
}

my_example <- example_funfun(",")

txt <- "1,2,3,4"
my_example$write(object = txt, file = tempfile(fileext = ".csv"))
#> Hi! The separation marker is: ','
#> We wrote the file out to:
#> /var/folders/9c/k3wqmhhx4qsb3fd66n4prhlw0000gq/T//Rtmp3p4VpA/file10446789624c4.csv
```

<sup>Created on 2024-03-16 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.3.3 (2024-02-29)
#>  os       macOS Sonoma 14.3.1
#>  system   aarch64, darwin20
#>  ui       X11
#>  language (EN)
#>  collate  en_US.UTF-8
#>  ctype    en_US.UTF-8
#>  tz       Australia/Hobart
#>  date     2024-03-16
#>  pandoc   3.1.1 @ /Applications/RStudio.app/Contents/Resources/app/quarto/bin/tools/ (via rmarkdown)
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version date (UTC) lib source
#>  bit           4.0.5   2022-11-15 [1] CRAN (R 4.3.0)
#>  bit64         4.0.5   2020-08-30 [1] CRAN (R 4.3.0)
#>  cli           3.6.2   2023-12-11 [1] CRAN (R 4.3.1)
#>  crayon        1.5.2   2022-09-29 [1] CRAN (R 4.3.0)
#>  digest        0.6.34  2024-01-11 [1] CRAN (R 4.3.1)
#>  evaluate      0.23    2023-11-01 [1] CRAN (R 4.3.1)
#>  fansi         1.0.6   2023-12-08 [1] CRAN (R 4.3.1)
#>  fastmap       1.1.1   2023-02-24 [1] CRAN (R 4.3.0)
#>  fs            1.6.3   2023-07-20 [1] CRAN (R 4.3.0)
#>  glue          1.7.0   2024-01-09 [1] CRAN (R 4.3.1)
#>  hms           1.1.3   2023-03-21 [1] CRAN (R 4.3.0)
#>  htmltools     0.5.7   2023-11-03 [1] CRAN (R 4.3.1)
#>  knitr         1.45    2023-10-30 [1] CRAN (R 4.3.1)
#>  lifecycle     1.0.4   2023-11-07 [1] CRAN (R 4.3.1)
#>  magrittr      2.0.3   2022-03-30 [1] CRAN (R 4.3.0)
#>  pillar        1.9.0   2023-03-22 [1] CRAN (R 4.3.0)
#>  pkgconfig     2.0.3   2019-09-22 [1] CRAN (R 4.3.0)
#>  purrr         1.0.2   2023-08-10 [1] CRAN (R 4.3.0)
#>  R.cache       0.16.0  2022-07-21 [2] CRAN (R 4.3.0)
#>  R.methodsS3   1.8.2   2022-06-13 [2] CRAN (R 4.3.0)
#>  R.oo          1.26.0  2024-01-24 [2] CRAN (R 4.3.1)
#>  R.utils       2.12.3  2023-11-18 [2] CRAN (R 4.3.1)
#>  R6            2.5.1   2021-08-19 [1] CRAN (R 4.3.0)
#>  readr         2.1.5   2024-01-10 [1] CRAN (R 4.3.1)
#>  reprex        2.1.0   2024-01-11 [2] CRAN (R 4.3.1)
#>  rlang         1.1.3   2024-01-10 [1] CRAN (R 4.3.1)
#>  rmarkdown     2.25    2023-09-18 [1] CRAN (R 4.3.1)
#>  rstudioapi    0.15.0  2023-07-07 [1] CRAN (R 4.3.0)
#>  sessioninfo   1.2.2   2021-12-06 [2] CRAN (R 4.3.0)
#>  styler        1.10.2  2023-08-29 [2] CRAN (R 4.3.0)
#>  tibble        3.2.1   2023-03-20 [1] CRAN (R 4.3.0)
#>  tidyselect    1.2.0   2022-10-10 [1] CRAN (R 4.3.0)
#>  tzdb          0.4.0   2023-05-12 [1] CRAN (R 4.3.0)
#>  utf8          1.2.4   2023-10-22 [1] CRAN (R 4.3.1)
#>  vctrs         0.6.5   2023-12-01 [1] CRAN (R 4.3.1)
#>  vroom         1.6.5   2023-12-05 [1] CRAN (R 4.3.1)
#>  withr         3.0.0   2024-01-16 [1] CRAN (R 4.3.1)
#>  xfun          0.42    2024-02-08 [1] CRAN (R 4.3.1)
#>  yaml          2.3.8   2023-12-11 [1] CRAN (R 4.3.1)
#> 
#>  [1] /Users/nick/Library/R/arm64/4.3/library
#>  [2] /Library/Frameworks/R.framework/Versions/4.3-arm64/Resources/library
#> 
#> ──────────────────────────────────────────────────────────────────────────────
```

</details>

Or is there some deeper stuff happening with targets here?

brownag: I don't think that will work, or at least I tried it before arriving at the body<- solution I proposed in #11 [here](https://github.com/njtierney/geotargets/issues/11#issuecomment-1990694724). Tried doing it again now because I hoped I had missed something. 


What you propose works fine interactively in your global environment, or in a reprex, etc. but the context where those `tar_format()` functions get evaluated you will not have the e.g. `sep_type` object defined. 

Essentially the functions `tar_format()` uses need to be self-contained, not relying on the prior parent frame. You actually want `sep_type` to be evaluated sooner as it is when we modify the function body.

If you implement your suggestion for `tar_terra_rast()` you will get something like this:

```
Loading required namespace: terra
▶ dispatched target test_terra_rast
✖ errored target test_terra_rast
✖ errored pipeline [0.13 seconds]
Error:
! Error running targets::tar_make()
Last error message:
    _store_ object 'filetype' not found
Last error traceback:
    No traceback available.
```



<details>


```r
# tar-terra-rast.R: create_format_terra_raster
#...
 .myfun <- function(filetype, gdal) {
        .write_terra_raster <- function(object, path) {
            terra::writeRaster(
                object,
                path,
                filetype = filetype,
                overwrite = TRUE,
                gdal = gdal, 
            )
        }
        return(list(write = .write_terra_raster))
    }

    targets::tar_format(
        read = function(path) terra::rast(path),
        write = .myfun(filetype, gdal)$write,
        marshal = function(object) terra::wrap(object),
        unmarshal = function(object) terra::unwrap(object)
    )
# ...
```
</details> 

Some other ideas I tried were using `purrr::partial()` to compose functions with some arguments pre-specified and couldn't get that working. Also forcing evaluation and injecting values with {rlang} metaprogramming operators also did not appear to work, but that might be worth re-investigating now.

njtierney: Gotcha, thanks for explaining that! It looks like we lose all reference to the parent frames here, I guess it might be evaluated in a separate R process, so making changes directly to the body of the function makes sense. I'm surprised that metaprogramming things from {rlang} don't work out, it feels like this is the sort of thing where you want to change some aspect of the AST or something. But effectively, I think, that's what you're doing with the clever `body()<-` calls.

If it continues to be a problem we can re-evaluate some possible solutions with rlang - I might even tag in Will into the issues to see if he has had this problem?

brownag:
We now have a better way to do this, that does not require body<- nor getting package options inside write function body

PR #40, opened 2024-03-16
fix for empty GEOTARGETS_GDAL_RASTER_CREATION_OPTIONS

brownag:
In #30 I replaced the `unset` value returned by `Sys.getenv()` for GEOTARGETS_GDAL_RASTER_CREATION_OPTIONS

Calling `strsplit("", ";)[[1]]` returns `character(0)`, which will cause an error in `Sys.getenv()`. This PR gets us the intended default value of `""`


njtierney:
Nice work! That's a good catch. I think writing a test here would be a good idea - lodged #41

PR #43, opened 2024-03-19
`tar_format()` functions: Use `substitute()` into function body rather than `body<-`

### LINKED ISSUE #38: Apparent bugs / problems with function body modification with covr+testthat+tar_test()

brownag:
- Uses a simpler approach to updating the read/write function templates used in format creation methods, thanks to Will Landau for suggestion
- Fixes #38 by avoiding the problematic syntax, normal tests and tests in covr now work

### DIFF
--- a/R/tar-terra-rast.R
+++ b/R/tar-terra-rast.R
@@ -109,21 +109,15 @@ create_format_terra_raster <- function(filetype, gdal, ...) {
 
     gdal <- gdal %||% geotargets_option_get("gdal.raster.creation_options")
 
-    # NOTE: Option getting functions are set in the .write_terra_raster function template
-    #       to resolve issue with body<- not working in some evaluation contexts ({covr}).
-    # TODO: It should be fine to have filetype and gdal as NULL
-    .write_terra_raster <- function(object, path) {
+    .write_terra_raster <- eval(substitute(function(object, path) {
         terra::writeRaster(
             object,
             path,
-            filetype = geotargets::geotargets_option_get("gdal.raster.driver"),
+            filetype = filetype,
             overwrite = TRUE,
-            gdal = geotargets::geotargets_option_get("gdal.raster.creation_options")
+            gdal = gdal
         )
-    }
-
-    body(.write_terra_raster)[[2]][["filetype"]] <- filetype
-    body(.write_terra_raster)[[2]][["gdal"]] <- gdal
+    }, list(filetype = filetype, gdal = gdal)))
 
     targets::tar_format(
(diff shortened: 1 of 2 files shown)


njtierney:
Ah, fantastic!

Issue #34, opened 2024-03-15
Adopt API of `tar_options_set` for `geotargets_options_set`


Aariq:
`targets::tar_options_set()` just uses named args with NULL as default and it might be nice for users to have a similar experience with `geotargets_options_set()`.  E.g.

```r 
geotargets_options_set <- function(
  gdal_raster_driver = NULL,
  gdal_raster_creation_options = NULL,
  gdal_vector_driver = NULL,
  gdal_vector_creation_options = NULL
  ){ ...
 
```

And because we are setting these options with defaults on package load, we can do arguments to `tar_*()` functions similar to `targets` like so:

```r 
tar_terra_rast <- function(name,
                           command,
                           pattern = NULL,
                           filetype = geotargets::geotargets_options_get("gdal_raster_driver"),
                           gdal = geotargets::geotargets_options_get("gdal_raster_creation_options"),
                           ...,
                           tidy_eval = targets::tar_option_get("tidy_eval"),


```

And remove `geotargets_options_get()` from the body of these functions.

brownag:
I really like this idea, I have already been bitten by typos and mistakes with these. Being able to use autocomplete and documentation of function parameters/usage would make it _much_ clearer.

I also like the idea to get the options in the function argument defaults. This could remove most usages, except in the internal functions we pass to `tar_format()` (until we figure out the covr body<- issue)

Aariq:
I also think we might want to switch the precedence of env variables and options() to have options override env variables since .Renviron is only read in on the start of an R session.

So, something like

``` r
geotargets_gdal_raster_creation_options <- getOption("geotargets.gdal.raster.creation_options", default = Sys.getenv("GEOTARGETS_GDAL_RASTER_CREATION_OPTIONS", unset = "ENCODING=UTF-8"))
```

Aariq:
I asked for some opinions of how to do this in rOpenSci Slack and got two options that both seem good.

1) from Hao Ye & Jon Harmon: on package load, set default options, but only the ones that haven't been set already (via options() or env vars).  E.g.: https://github.com/c4r-io/gdrive-automation/blob/c3ce3e5a9963b36ccc480243e96b5ab65eb12872/R/gdrive.automation-package.R#L12

2) from Gábor Csárdi: don't set any options at load time (neither defaults nor from env vars) and just include the defaults in functions (like in PR #45)

[...]

PR #45, opened 2024-03-20
Options makeover

Aariq:
Refactor to address #34. I've taken the approach of not setting any options on package load, but rather storing defaults for individual options in the relevant functions. 

This PR does two main things:

1. Refactors how we create write functions to take advantage of the new `tar_resources_custom_format()` function in the development version of `targets` to pass options to write functions in custom formats.  
    - This provides a cleaner alternative to `eval(substitute())` and works with distributed `crew` workers. 
    - see [example](https://github.com/ropensci/targets/discussions/1263#discussioncomment-9063458)
    - Unfortunately, requires `targets` to be installed from GitHub for now.  Eventually will change version requirement to >=1.6.1 and remove the Remotes field from DESCRIPTION.
2. Simplifies the way `geotargets_option_get()` and `geotargets_option_set()` work.  
   - `geotargets_option_set()` is now just a wrapper around `options()` with some guardrails. 
   - `geotargets_options_get()` looks in `getOption()` and then in `Sys.getenv()` for options, otherwise returns NULL.  It also checks if the option name is valid with `rlang::arg_match0()`
   - Defaults options are not set.  Defaults are stored in `tar_*()` functions


Aariq:
Just realized this is definitely not ideal as any geotargets options set in .Rprofile will be overridden on package load.  I like the idea of storing geotargets options in a separate environment that is initialized on package load, I just don't understand how to do it (yet).

njtierney:
I'm wondering if we could maybe use some of the `withr::local_options()` type things to test how these options work? The testthat ["test figures" vignette](https://testthat.r-lib.org/articles/test-fixtures.html) provides some nice examples here.

I'm happy for you to merge this currently if you're happy with it and we can add a separate issue about adding more comprehensive tests for setting/getting options? Also happy to start the work on that issue if you'd like a break from it :)

brownag (approved):
> I've taken the approach of not setting any options on package load, but rather storing defaults for individual options in the relevant functions. 

I think the package should not set options on load so this is a good decision IMO.

>  If no options are set by the user, it falls back on defaults stored within each `tar_*()` function.

In #19 I wondered whether it would be easier to maintain if they were all set in the options get function, but this is probably a better solution: there may be special cases handled in specific `tar_*()` functions that can't be handled more generally

Aariq:
> I'm happy for you to merge this currently if you're happy with it and we can add a separate issue about adding more comprehensive tests for setting/getting options? Also happy to start the work on that issue if you'd like a break from it :)

No, I think I'll add tests before this gets merged.  I want to make sure that options set globally actually get used correctly in a targets pipeline.  

I also keep going back and forth on whether running `geotargets_options_set()` with no arguments should do anything (or if we even need that function).  Perhaps the logic in `geotargets_options_set()` (user arg > options > .Renviron) should just be in each `tar_*()` function?

Aariq:
> In #19 I wondered whether it would be easier to maintain if they were all set in the options get function, but this is probably a better solution: there may be special cases handled in specific `tar_*()` functions that can't be handled more generally

I keep going back and forth on this.  After taking a long break from this PR, I'm now thinking it makes more sense for `tar_options_get()` to handle the logic of user-supplied %||% getOption() %||% env var %||% default (which, I realize, is basically how it works currently). 

Also, using `options()` doesn't work for pipelines with multiple workers 🤦‍♂️. 

Going to study how options are done in `targets` some more and come back to this.

Aariq:
Edited the PR description up top and now I think this is ready to review

Aariq:
Tests should start passing again once `targets` v 1.7.0 hits CRAN

njtierney (approved):
This all looks really good to me! I will update the changes from #50 in here now, and update the tests etc. But this looks like it simplifies things greatly.

Issue #22, opened 2024-03-14
using `geodata::gadm` with geotargets


njtierney:
`geodata::gadm` is commonly used to get shapefiles for countries at various levels of boundaries.

This issue is to talk about how to use gadm within geotargets - related to #5.

Some code to download a country shapefile might be something like:

```r
get_gadm_country <- function(country = "Australia") {
    dir.create("data/shapefiles", recursive = TRUE, showWarnings = FALSE)
    geodata::gadm(
        country = country,
        level = 0,
        path = "data/shapefiles",
        version = "4.1",
        # low res
        resolution = 2
    )
}
```

My question is how should we recommend that users implement and use something like this in a targets workflow? At this stage I haven't tested this code out with the latest version of {geotargets} so it might all just work, but my specific concerns are around:

- What do we recommend users do when creating functions like this? Do they need to create directories?
- Should we provide helpers so users can do this kind of data retrieval easily?

Also porting in @brownag 's comment from #5 in here as it is relevant:

> In the case of `get_gadm_country()` you will get one .rds file in the directory path for each country specified. Each .rds file contains a wrapped SpatVector of a single country. When the cache exists for a particular country/version/resolution, rather than calling download.file() the SpatVector is read from file and unwrapped. Each of these could be "file" targets if you can determine what a particular call to gadm() should produce (i.e. 1 file with specific file name given parameters in path for each country specified)

> After https://github.com/Aariq/geotargets/commit/f82edd47ababf7b073f388e87d4d5806cc747132 you can easily do something like tar_terra_vect(some_countries, get_gadm_country(c("Australia", "New Zealand"))) which will write the SpatVector combination of two .rds files as a new seek-optimized shapefile (.shz). If the source .rds files don't exist (i.e. on first run of the pipeline) or are deleted (i.e. a temp folder is used), they would be re-downloaded. This does not rely on external files outside the target store, but technically you have copies of the data in the initial gadm() download (stored as PackedSpatVector) and the target(s) (stored as zipped shapefile)

> If you want to have the file(s) produced by gadm() be the targets... we need to consider that they are not shapefiles but rather PackedSpatVector in "rds" format... and a single call could produce multiple "rds" files (multiple targets)

[...]

PR #55, opened 2024-04-19
'getting started' vignette

### LINKED ISSUE #17: Create vignette for tar_terra_* targets use
See https://github.com/ropensci-books/targets/blob/main/walkthrough.qmd and https://github.com/ropensci/stantargets/blob/main/vignettes/introduction.Rmd for details on setting up targets in a vignette

njtierney:
Still need to fill this in with more content from issues.

For example

- using gdam and vsizip and cgaz https://github.com/njtierney/geotargets/issues/22
- using `geotargets_option_set` - https://github.com/njtierney/geotargets/pull/45

This will resolve #17


vignettes/using-terra.Rmd
```diff
@@ -0,0 +1,126 @@
...
+      terra::project(
+        terra::rast(system.file("ex", "elev.tif", package = "terra")) * z,
+        projection
+      )
+    }
+    list(
+      tar_terra_sprc(
+        raster_elevs,
+        # two rasters, one unaltered, one scaled by factor of 2 and
+        # reprojected to interrupted good homolosine
```
Aariq, line 115:
```suggestion
        # reprojected to interrupted Goode homolosine
```

vignettes/using-terra.Rmd
```diff
@@ -0,0 +1,126 @@
...
+    }
+    list(
+      tar_terra_vect(
+        terra_vect_example,
+        lux_area()
+      )
+    )
+  })
+  tar_make()
+  x <- tar_read(terra_rast_example)
```
Aariq, line 91:
```suggestion
  x <- tar_read(terra_vect_example)
```

vignettes/using-terra.Rmd
```diff
@@ -0,0 +1,126 @@
...
+        command = terra::sprc(list(
+          elev_scale(1),
+          elev_scale(2, "+proj=igh")
+        ))
+      )
+    )
+  })
+  targets::tar_make()
+  x <- targets::tar_read(raster_elevs)
+})
```
Aariq, line 125:
```suggestion
  x <- targets::tar_read(raster_elevs)
  x
})
```

Aariq (commented):
It's a good start!  I think eventually we should add an example of how to change file type and pass GDAL options, but that could be in a separate PR if you'd rather.

Aariq:
Not sure why, but when I `build_vignette()` and then view it with `vignette("using-terra")`, I don't see any output, just code.

njtierney:
Good idea with precomputing!

Aariq:
Because of the trick to set `root.dir` to a temporary directory, you don't need `tar_dir()` at all I think.  All examples have three chunks: one with echo: false to write the _targets.R file with `tar_script()`, one with eval: false to show the contents of _targets.R, and one with echo and eval true to run the pipeline and inspect results.

I didn't include much about the raster and vector GDAL options, because I don't have a good motivating example, but I did include examples of changing the driver and added an example with `preserve_metadata =  "zip"`.

Aariq:
@njtierney, I think this is pretty much done.  Can you take a look when you get a chance?

njtierney:
Thanks so much for this, @Aariq !

I think it's great. I think that down the track we could provide some more information about using things like `pattern = map()`, but that might be best put in another future vignette, or perhaps link to a repository that contains a worked example.

Happy to merge!

Aariq:
Yeah, the plan was to show some dynamic branching examples in a separate vignette focused on `tar_terra_tile()` (#111), but I'm stuck on that because I really wanted `tile_blocksize()` to work (#99)!

PR #57, opened 2024-04-24
Add tests with multiple `crew` workers

Aariq:
Adds tests with multiple workers to ensure that marshaling / unmarshaling works.  Adds `crew` as a Suggests dependency.


njtierney (approved):
Awesome, looks great!

Issue #58, opened 2024-04-25
`tar_terra_rast()` doesn't save units or varnames


Aariq:
I suspect this is because those metadata are stored in a "sidecar file" when `filetype = "GeoTIFF"`, but I haven't investigated further.

``` r
library(targets)
tar_script({
    make_rast <- function() {
        x <- terra::rast(system.file("ex/elev.tif", package = "terra"))
        terra::units(x) <- "m"
        terra::varnames(x) <- "elev"
        x
    }
    list(
        geotargets::tar_terra_rast(
            rast,
            make_rast()
        )
    )
})
tar_make()
#> ▶ dispatched target rast
#> ● completed target rast [0.017 seconds]
#> ▶ ended pipeline [0.148 seconds]
tar_load(rast)
terra::units(rast)
#> [1] ""
terra::varnames(rast)
#> [1] "rast"
terra::names(rast)
#> [1] "elevation"
```

<sup>Created on 2024-04-25 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Aariq:
Yeah, looks like units are store in an aux.json file.  `varnames()` aren't read back in by `rast()` so I guess that one isn't our problem.  Related to #37

Aariq:
I've tried a half dozen or so differente GDAL drivers and they all create an aux.json file when a SpatRaster has units. Ideas:
1) always zip files after writing and us vsizip to read back in
2) have `tar_terra_rast(my_map)` create two targets: `my_map_rast` and `my_map_meta`.  Not sure how reading this back in would work though

Aariq:
Found this: https://github.com/rspatial/terra/blob/master/NEWS.md#new-18
aux.json file is written for all drivers except when using `writeCDF()`

njtierney:
Thanks for this! I'll take a look later this week on my geotargets day.

amart90:
Hello folks! Thank you for the work on this package. I've been interested in it for a while and am now finally getting a chance to use it. I noticed that `geotargets::tar_terra_rast()` seems to be stripping the time attribute of my raster and I'm thinking that it is probably related to this. Here is a reproducible example:

```r
library(targets)
tar_script({
  make_rast <- function() {
    x <- terra::rast(system.file("ex/elev.tif", package = "terra"))
    terra::time(x) <- as.Date("2010-01-01")
    x
  }
  list(
    geotargets::tar_terra_rast(
      rast,
      make_rast()
    )
  )
})

[...]

PR #63, opened 2024-04-29
[draft] tar_terra_rast_wrap: multi-target method to preserve SpatRaster metadata

brownag:
This is a draft PR that might be able to address #58 

This is a completely different way of managing target files--where the target file in _targets/objects/ is an RDS file (like ordinary targets) containing a PackedSpatRaster which is backed by a cached geospatial data file (and any sidecars) held in a user-specified folder

 - Source data file is written using `terra::wrapCache()` to a user-specified cache directory
 - Target (custom format, saves RDS in target store) is created for PackedSpatRaster which is linked to the cache files
 - Target (format="file") is created for cache files
 - Add cache-managing functions `geotargets_destroy_cache()`, `geotargets_init_cache()` and env option for cache path GEOTARGETS_CACHE_DIR (and associated methods)

For now this only works for SpatRaster, but I think a similar solution could be developed for SpatVectorProxy (although this would require either changes to `wrapCache()` in terra, or a custom `wrapCache()`-like method developed for this case)

Current "issue" is that you can modify the cache (intentionally or unintentionally) and the main target will not be invalidated. I tried tracking the cache directory before running the caching target, but then this leads to the caching having to run twice before it is skipped.

Example of storing units and categories:

``` r
library(targets)
tar_script({
    
    make_rast1 <- function() {
        x <- terra::rast(system.file("ex/elev.tif", package = "terra"))
        terra::units(x) <- "m"
        terra::varnames(x) <- "elev"
        x
    }
    
    make_rast2 <- function() {
        x <- terra::rast(system.file("ex/elev.tif", package = "terra"))
        y <- terra::classify(x, cbind(c(0, 300, 500),
                                      c(300, 500, 1000),
                                      1:3))
        levels(y) <- data.frame(value = 1:3,
                                category = c("low", "med", "hi"))
        y
    }
    
    list(
        geotargets::tar_terra_rast_wrap(
            rast1,
            make_rast1()
        ),
        geotargets::tar_terra_rast_wrap(
            rast2,
            make_rast2()
        )
    )
})

tar_make()
#> ▶ dispatched target rast1
#> ● completed target rast1 [0.009 seconds]
#> ▶ dispatched target rast2
#> ● completed target rast2 [0.016 seconds]
#> ▶ dispatched target rast1_cache_files
#> ● completed target rast1_cache_files [0 seconds]
#> ▶ dispatched target rast2_cache_files
#> ● completed target rast2_cache_files [0 seconds]
#> ▶ ended pipeline [0.334 seconds]

x_raw <- readRDS("_targets/objects/rast1")
x <- tar_read(rast1)

x_raw@attributes
#> $sources
#>   sid
#> 1   1
#>                                                                             source
#> 1 /tmp/RtmpfZAMV0/reprex-67d786408f708-bared-pika/geotargets_cache/rast1/rast1.tif
#>   bands nlyr
#> 1     1    1
#> 
#> $units
#> [1] "m"

terra::units(x)
#> [1] "m"

# varnames not preserved in PackedSpatRaster either
terra::varnames(x)
#> [1] "rast1"

x
#> class       : SpatRaster 
#> dimensions  : 90, 95, 1  (nrow, ncol, nlyr)
#> resolution  : 0.008333333, 0.008333333  (x, y)
#> extent      : 5.741667, 6.533333, 49.44167, 50.19167  (xmin, xmax, ymin, ymax)
#> coord. ref. : lon/lat WGS 84 (EPSG:4326) 
#> source      : rast1.tif 
#> name        : elevation 
#> min value   :       141 
#> max value   :       547 
#> unit        :         m

x <- tar_read(rast2)

terra::levels(x)
#> [[1]]
#>   value category
#> 1     1      low
#> 2     2      med
#> 3     3       hi

x
#> class       : SpatRaster 
#> dimensions  : 90, 95, 1  (nrow, ncol, nlyr)
#> resolution  : 0.008333333, 0.008333333  (x, y)
#> extent      : 5.741667, 6.533333, 49.44167, 50.19167  (xmin, xmax, ymin, ymax)
#> coord. ref. : lon/lat WGS 84 (EPSG:4326) 
#> source      : rast2.tif 
#> categories  : category 
#> name        : category 
#> min value   :      low 
#> max value   :       hi

tar_read(rast1_cache_files)
#> [1] "geotargets_cache/rast1/rast1.tif"         
#> [2] "geotargets_cache/rast1/rast1.tif.aux.json"

# all skip
tar_make()
#> ✔ skipped target rast1
#> ✔ skipped target rast2
#> ✔ skipped target rast1_cache_files
#> ✔ skipped target rast2_cache_files
#> ✔ skipped pipeline [0.119 seconds]

# change the rast1 cache by changing units
x <- jsonlite::read_json("geotargets_cache/rast1/rast1.tif.aux.json")
x[[1]][[1]] <- "km" 
jsonlite::write_json(x, "geotargets_cache/rast1/rast1.tif.aux.json")

# need to rebuild rast1 target
tar_make()
#> ✔ skipped target rast1
#> ✔ skipped target rast2
#> ▶ dispatched target rast1_cache_files
#> ● completed target rast1_cache_files [0.001 seconds]
#> ✔ skipped target rast2_cache_files
#> ▶ ended pipeline [0.153 seconds]

# all skip
tar_make()
#> ✔ skipped target rast1
#> ✔ skipped target rast2
#> ✔ skipped target rast1_cache_files
#> ✔ skipped target rast2_cache_files
#> ✔ skipped pipeline [0.118 seconds]
```


R/tar-terra-wrap.R
```diff
@@ -0,0 +1,153 @@
...
+
+    rast_cache_files <- targets::tar_target_raw(
+        paste0(name, "_cache_files"),
+        str2expression(paste0("
+            list.files(
+                file.path(", shQuote(cachedir), ", ", shQuote(name),"),
+                full.names = TRUE,
+                recursive = TRUE
+            )")),
+        format = "file_fast",
```
Aariq, line 146: You could allow the option of either "file" or "file_fast" here, but I'm guessing it doesn't really matter since it seems like nothing will ever depend on this target.

R/tar-terra-wrap.R
```diff
@@ -0,0 +1,153 @@
...
+        deps = name
+    )
```
Aariq, line 148: I'm questioning whether this target even needs to exist. Unless there is a way to make it be "upstream" of the `wrapCache` target, then it doesn't really serve a purpose.  Invalidating this target will never do anything, and there's no reason to use this target rather than the upstream one in a pipeline.  So maybe this doesn't need to return multiple targets.

Aariq (commented):
Sorry it has taken so long to get to this.  I'm excited about this PR because it seems to me like this could actually make sense to just be the default way that `geotargets` works with `terra` and that would solve a lot of issues.

R/geotargets-cache.R
```diff
@@ -0,0 +1,47 @@
...
+#' geotargets_destroy_cache()
+#'
+#' }
+geotargets_destroy_cache <- function(name = NULL, init = FALSE) {
+    cachedir <- geotargets_option_get("cache.dir")
+    target_cache_dir <- file.path(cachedir %||% "geotargets_cache", name %||% "")
+    res <- unlink(target_cache_dir, recursive = TRUE)
+    if (init) geotargets_init_cache(name = name)
+    invisible(res)
+}
```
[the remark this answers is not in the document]
Aariq: Actually, now that I think of it, the directory names inside the cache are target names, yeah?  So if this could  get all those dir names and pass them to `tar_invalidate()` I think it would make this function a lot more useful.

R/tar-terra-wrap.R
```diff
@@ -0,0 +1,153 @@
...
+        envir = envir,
+        tidy_eval = tidy_eval
+    )
+
+    .format_terra_rast_wrap_write <- eval(substitute(function(object, path) {
+        # TODO: provide mapping of major file types to extensions
+        extension <- switch(filetype,
+                            "GTiff" = ".tif",
+                            "GPKG" = ".gpkg",
+                            "")
```
Aariq, line 91: Is this necessary?  We don't need file extensions when writing to `_targets/`, why do we need them here?

R/tar-terra-wrap.R
brownag: Agreed, I think it would be ideal to do it in one target. You are right that as implemented it is not doing what I intended.

R/tar-terra-wrap.R
brownag: It is not necessary for this `wrapCache()` approach in general. 

I think I was thinking that the files in _geotargets cache folder might be something that could be accessed directly outside of R e.g. in other GIS software where having the extension would be helpful. Probably we don't want to encourage that type of use as it could change data or metadata outside the targets pipeline, so it is probably best to remove all handling of file extensions.

Aariq:
I'm going to make a branch based on this one to test out making this wrapCache behavior an *option* rather than another `tar_terra_*()` function.  I think recent changes in `targets` might make it simpler to have more flexible behavior in `tar_format()` that will allow this.

Issue #61, opened 2024-04-29
split_semicolon() function


njtierney:
We do the pattern:

```r
strsplit(options, ";")[[1]]
```

and

```r
paste0(options, collapse = ";")
```

A bit in the package. A function, `semicolon_split(option)` and converse I think is worthwhile.

``` r
semicolon_split <- function(options){
    strsplit(options, ";")[[1]]
}

some_options <- c("THING;ANOTHER;THINGY")

semicolon_split(some_options)
#> [1] "THING"   "ANOTHER" "THINGY"

semicolon_paste <- function(vec){
    paste0(vec, collapse = ";")
}

some_options <- c("THING;ANOTHER;THINGY")

some_options
#> [1] "THING;ANOTHER;THINGY"

semicolon_split(some_options)
#> [1] "THING"   "ANOTHER" "THINGY"

semicolon_split(some_options) |> semicolon_paste()
#> [1] "THING;ANOTHER;THINGY"
```

<sup>Created on 2024-04-29 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.3.3 (2024-02-29)
#>  os       macOS Sonoma 14.3.1
#>  system   aarch64, darwin20
#>  ui       X11
#>  language (EN)
#>  collate  en_US.UTF-8
#>  ctype    en_US.UTF-8
#>  tz       Australia/Hobart
#>  date     2024-04-29
#>  pandoc   3.1.1 @ /Applications/RStudio.app/Contents/Resources/app/quarto/bin/tools/ (via rmarkdown)
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version date (UTC) lib source
#>  cli           3.6.2   2023-12-11 [1] CRAN (R 4.3.1)
#>  digest        0.6.35  2024-03-11 [1] CRAN (R 4.3.1)
#>  evaluate      0.23    2023-11-01 [1] CRAN (R 4.3.1)
#>  fastmap       1.1.1   2023-02-24 [1] CRAN (R 4.3.0)
#>  fs            1.6.3   2023-07-20 [1] CRAN (R 4.3.0)
#>  glue          1.7.0   2024-01-09 [1] CRAN (R 4.3.1)
[...]

PR #65, opened 2024-04-29
`tar_terra_sprc()` simplification

Aariq:
Removes some repeated or unnecessary code I noticed in `tar_terra_sprc()`.  Implements suggestion in #61.

### DIFF
--- a/R/geotargets-option.R
+++ b/R/geotargets-option.R
@@ -37,11 +37,16 @@ geotargets_option_set <- function(
         gdal_vector_creation_options = NULL
 ) {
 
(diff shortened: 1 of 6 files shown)


Aariq:
I'm not sure `semicolon_split()` and `semicolon_paste()` were worth it because it seems difficult to use unexported functions inside read/write functions of custom formats.

njtierney:
I feel you re the semicolon_paste/split functions - they feel like they should be useful but I can understand that in this context they don't really help!

njtierney (approved):
This all looks good to me, happy to merge when you like :)

PR #70, opened 2024-05-15
Release 0.1.0

### LINKED ISSUE #68: Release geotargets 0.1.0
First release (for github):

* [x] `usethis::use_news_md()`
* [x] Update (aspirational) install instructions in README
* [x] Proofread `Title:` and `Description:`
* [x] Check that all exported functions have `@return` and `@examples`
* [x] Check [licensing of included files](https://r-pkgs.org/license.html#sec-code-you-bundle)
* [x] Review <https://github.com/DavisVaughan/extrachecks>

Prepare for release:

* [x] `git pull`
* [x] `urlchecker::url_check()`
* [x] `devtools::build_readme()`
* [x] `devtools::check(remote = TRUE, manual = TRUE)`
* [x] `devtools::check_win_devel()`
* [x] `git push`
* [x] Draft blog post
* [x] Finish & publish blog post
* [ ] `usethis::use_github_release()`
* [ ] `usethis::use_dev_version(push = TRUE)`
* [ ] `usethis::use_news_md()`
* [ ] Toot

njtierney:
Resolves #68

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: geotargets
 Title: Targets extensions for geospatial formats
-Version: 0.0.0.9000
+Version: 0.1.0
 Authors@R: c(
     person(
       given = "Nicholas", 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# geotargets 0.0.0.9000 (development version)
+# geotargets 0.1.0 (14 May 2024)
 
 * Created `tar_terra_rast()` and `tar_terra_vect()` for targets that create `SpatRaster` and `SpatVector` objects, respectively
 * Created `tar_terra_sprc()` that creates a `SpatRasterCollection` object.
--- a/R/geotargets-option.R
+++ b/R/geotargets-option.R
@@ -27,9 +27,32 @@
 #' `geotargets_options_set(gdal_raster_driver = "GTiff")` is equivalent to
 #' `options("geotargets.gdal.raster.driver" = "GTiff")`.
 #'
+#' @return Specific options, such as "gdal.raster.driver". See "Details" for
+#'   more information.
+#'
 #'
 #' @rdname geotargets-options
 #' @export
+#' @examples
+#' if (Sys.getenv("TAR_LONG_EXAMPLES") == "true") {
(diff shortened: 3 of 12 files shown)


Aariq (approved):

Aariq:
I couldn't reproduce the error on r-devel with rocker/geospatial:devel.  However, I got a different test error from that container:

```
Error (test-tar-terra.R:62:5): tar_terra_vect() works
<tar_condition_run/tar_condition_targets/rlang_error/error/condition>
Error: Error running targets::tar_make()
Last error message:
    the write() function in tar_format() must not create a directory. Found directories inside the data store where there should only be files: _targets/objects/test_terra_vect_shz

```

I wonder if this has something to do with differences in GDAL versions? `terra::gdal()` reports `"3.0.4"` on this container.

njtierney:
That is a bit strange re the crew workers! Thanks for checking that out in a docker container. 

3.0.4 is [4 years old](https://github.com/OSGeo/gdal/releases/tag/v3.0.4), it makes me think that maybe we should set a minimum version GDAL version in the DESCRPTION under SystemReqs and then also maybe manage that in GH actions, where we can provide a range of GDAL versions

brownag:
> I wonder if this has something to do with differences in GDAL versions? `terra::gdal()` reports `"3.0.4"` on this container.

Can confirm writing SOZip shapefile requires GDAL >= 3.7. Not sure if we should have a fallback to using utils::zip() (non-SOZip) in this case, or whether it should be an error/test should be skipped on old GDAL

Aariq:
Still looking into the error with the crew containers.  I'm going to see if I can set up the controller to save worker logs in a way that'll be accessible as an artifact from GitHub actions—that should be useful going forward.  

GDAL 3.7 was released only about a year ago, so it seems maybe not ideal to require >= 3.7, so it might be a good idea to fallback to `utils::zip()` with older versions. I can open an issue to discuss this.

Aariq:
Hmm, I'm stumped.  Nothing should have changed with those crew tests in this PR!  It's hard to debug because `targets::tar_test()` does too good of a job isolating things and I can't figure out how to print the crew worker logs on GitHub actions.

Aariq:
Well, looks like whatever was going on was a temporary GitHub issue.

PR #72, opened 2024-05-17
Add a step to github action to print GDAL version

Aariq:
For easier debugging.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -44,6 +44,9 @@ jobs:
           extra-packages: any::rcmdcheck
           needs: check
 
+      - name: GDAL version
+        run: gdalinfo --version
+
       - uses: r-lib/actions/check-r-package@v2.8.3
         with:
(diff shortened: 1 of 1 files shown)


Aariq:
Oh interesting.  I guess `terra` uses some different `gdal` on macOS and windows (I was wondering why those tests worked without an explicit step to install system dependencies).  E.g. on my computer `gdalinfo --version` gives `GDAL 3.8.5, released 2024/04/02` but `terra::gdal()` gives `[1] "3.5.3"`

brownag:
> Oh interesting. I guess `terra` uses some different `gdal` on macOS and windows (I was wondering why those tests worked without an explicit step to install system dependencies). E.g. on my computer `gdalinfo --version` gives `GDAL 3.8.5, released 2024/04/02` but `terra::gdal()` gives `[1] "3.5.3"`

If installing a binary package from CRAN or r-universe or similar, default for windows and macOS, you get whatever version is currently bundled by the repository maintainers. 

If compiling from source it will use your local installed version.

PR #75, opened 2024-05-21
Add CITATION.cff

Aariq:
Adds a CITATION.cff file that will be recognized by GitHub and Zenodo.  Also adds a reminder to update it when using `usethis::use_release_issue()`


njtierney:
That's super cool with the `release_bullets()` function, I didn't know about that.

Issue #74, opened 2024-05-20
Supporting various split-apply-combine spatial workflows


Aariq:
So there are at least 4 different ways (yikes!) to do a split-apply-combine type workflow with spatial data just in `terra` that I've encountered. It may be beneficial to figure out a way to implement these workflows using `targets` branching in `geotargets`

1) SpatRasters or SpatVectors with multiple layers can be split with `[[` and recombined with `c()` or `rast()`/`vect()`.  Also can be iterated over with `lapply()` or coerced to a list with `as.list()`
2) SpatRasterCollections hold a list of SpatRasters, possibly with different extents, resolutions, and projections.  Can be split with `[` (but not `[[`) and combined with `terra::sprc()`. `lapply()` also works on them and can be coerced with `as.list()`. (see #53)
3) SpatRasterDatasets hold a list of SpatRasters with the same resolution and projection (sub-datasets).  Can be split with `[` or `[[` and re-combined with `terra::sds()`.  `lapply()` also works and can be coerced with `as.list()`.  `names` are lost in the process (I assume this is a bug: https://github.com/rspatial/terra/issues/1513). (see #59)
4) ✅ `makeTiles()` splits a raster into tiles that are saved to disk and returns a vector of file paths.  Can be opened individually, worked on, and re-combined with `merge(sprc(rast(<list of tiles as SpatRasters>)))`.  Alternatively, the files on disk can be opened with `vrt()` (see #69)
5) ✅ iterate over lists of SpatRasters or SpatVectors (e.g. with dynamic branching)

### Some open questions:

1. Which of these (if any) should we try to support better in `geotargets` (e.g. with target factories)?
2. Are there existing similar patterns we can borrow from other Targetopia packages? (knowing that we can't customize the behavior of the `iteration` argument to `tar_target_raw()`)
3. Will any of these be easier to do if we adopt an approach like #63? (I suspect that a `tar_terra_tiles()` would be easier like this)
4. Should we try to figure this stuff out soon, or later after we've added basic support for `sf` and `stars`?

Aariq:
A 5th way: just iterating over lists of rasters or vectors (#77).  

[...]

PR #76, opened 2024-05-22
Add taget factory for splitting raster into tiles

Aariq:
I took a shot at creating a `tar_terra_tiles()` which is a target factory that creates three targets:
1) an upstream target that takes a SpatRaster and a template to define the tiles (the `y` argument of `terra::makeTiles()`), splits the raster into tiles, saves them to disk, and returns a vector of file paths.
2) a file target that maps over the file paths returned by `makeTiles()` using a target with format = "file" or "file_fast"
3) a downstream target that maps over the file paths and reads them in as rasters using dynamic branching.

This downstream target is a pattern that is ready to use in further targets with `pattern = map()` and starts to address #74.

This is heavily inspired by `tarchetypes::tar_files()`

The main downside to this pattern is that it basically duplicates the tiles on disk—once saved in a directory defined by the `tiles_dir` argument and again in the targets store when those files are read back in.  Deleting those files would invalidate the files target, but not the upstream target that creates them, so I'm forced to have the upstream target always re-run with `cue = tar_cue("always")`.  This is not ideal as making tiles is often a computationally expensive step.  I've left `#TODO` comments with questions and aspirations about how this might work better and would love feedback.


Aariq:
Ok, it is "working" now (not passing check, but there is a test that should work now to give you an idea of what it does).  Here's what the manifest looks like:

``` r
library(targets)
tar_script({
    library(targets)
    library(geotargets)
    # devtools::load_all()
    library(terra)
    list(
        tar_target(
            my_file,
            system.file("ex/elev.tif", package="terra"),
            format = "file"
        ),
        tar_terra_rast(
            my_map,
            terra::rast(my_file)
        ),
        tar_terra_tiles(
            name = rast_split,
            raster = my_map,
            template = terra::rast(ncols = 2, nrows = 2),
            tiles_dir = "tiles_test"
        )
    )
})
targets::tar_manifest(callr_function = NULL)
#> terra 1.7.71
#> # A tibble: 5 × 3
#>   name             command                                               pattern
#>   <chr>            <chr>                                                 <chr>  
#> 1 my_file          "system.file(\"ex/elev.tif\", package = \"terra\")"   <NA>   
#> 2 my_map           "terra::rast(my_file)"                                <NA>   
#> 3 rast_split_tile  "geotargets:::make_tiles(raster = my_map, template =… <NA>   
#> 4 rast_split_files "rast_split_tile"                                     map(ra…
#> 5 rast_split       "rast(rast_split_files)"                              map(ra…
```

<sup>Created on 2024-05-24 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Aariq:
Just pushed a fairly major improvement:
- No longer requires writing out to files outside of `_targets/`
- Now creates just two targets: 
    - an upstream target that is a list of numeric vectors that can be coerced to SpatExtent objects
    - a downstream target that iterates over the list of numeric vectors using them to set the `window()` of the original raster
- Exports two new helper functions, mostly so the command in `tar_manifest()` looks pretty, but also they might be helpful on their own.

R/tar_terra_tiles.R
```diff
@@ -0,0 +1,261 @@
...
+#' @export
+#'
+#' @examples
+#' f <- system.file("ex/elev.tif", package="terra")
+#' r <- terra::rast(f)
+#' create_tile_exts(r, ncol = 2, nrow = 2)
+create_tile_exts <- function(raster, ncol, nrow) {
+    template <- terra::rast(terra::ext(raster), ncol = ncol, nrow = nrow, crs = terra::crs(raster))
+    tile_ext <- terra::getTileExtents(raster, template)
+    lapply(1:nrow(tile_ext), \(i) tile_ext[i,])
```
njtierney, line 260:
```suggestion
    lapply(seq_len(nrow(tile_ext)), \(i) tile_ext[i,])
```

tests/testthat/test-tar_terra_tiles.R
njtierney, line 1: Small thing, but all the tests use hyphens to separate the words, can this change to hyphen separated?

R/tar_terra_tiles.R
```diff
@@ -0,0 +1,261 @@
...
+        storage = storage,
+        retrieval = retrieval,
+        cue = cue,
+        description = description
+    )
+
+    list(windows, tiles)
+}
+
+#' Copy a raster with a window
```
njtierney, line 206:
```suggestion
#' Copy a raster within a window
```

R/tar_terra_tiles.R
```diff
@@ -0,0 +1,261 @@
...
+#'
+#' Create a new SpatRaster object based on an original SpatRaster and a window.
```
njtierney, line 208:
```suggestion
#' Create a new SpatRaster object as specified by a window (area of interest) over the original SpatRaster.
```

R/tar_terra_tiles.R
```diff
@@ -0,0 +1,261 @@
...
+#' This is a wrapper around [terra::window()] which, rather than modifying the
+#' SpatRaster in place, returns a new SpatRaster leaving the original unchanged.
+#'
+#' @param raster A SpatRaster object
+#' @param window A SpatExtent object
```
njtierney, line 213:
```suggestion
#' @param window A SpatExtent object. The window of interest.
```

njtierney (changes requested):
Just some small changes, this is looking really nice!

R/tar_terra_tiles.R
```diff
@@ -0,0 +1,261 @@
...
+#'
+#' @param raster a `SpatRaster` object to be split into tiles
+#' @param ncol ncol
+#' @param nrow nrow
+#' @param filetype character. File format expressed as GDAL driver names passed
+#'   to [terra::makeTiles()]
+#' @param gdal character. GDAL driver specific datasource creation options
+#'   passed to [terra::makeTiles()]
+#'
+#' @param ... additional arguments not yet used
```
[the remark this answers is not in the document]
Aariq: This was for consistency with other functions `...` arg and I think it is probably grammatically correct as is, although could be "additional arguments which are not yet used".  Either way, I'll keep it as-is for now and if we want to change it we can change it for all functions in a separate PR.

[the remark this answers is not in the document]
njtierney: Yup that makes sense!

njtierney (approved):
Looks good to me!

R/tar_terra_tiles.R
njtierney: Yup that reads great!

Issue #83, opened 2024-07-17
Some thoughts on specifying tile size


njtierney:
It seems there could be 3 practical ways to specify tile size

1. pixel/dimension (e.g., 2x2, 3x3) 
2. Block size - use the block size to create better/more efficient tile size (e.g., if you have a 512 block size for a giant raster, it might make best computational sense to stick to tiles that are multiples of block size, since those blocks will need to be read in anyway)
3. Number of tiles - I want 9 tiles, for example. This might cut against the principles of block size - e.g., if you split a raster into 9 tiles, but the blocksize ends up with 4 tiles, then you might not be getting decent gains in read/write and operations.

Aariq:
For option 2 we could default to using `terra::fileBlocksize()` if `ncol` and `nrow` aren't supplied to `tar_terra_tiles()`.  We could also replace `ncol` and `nrow` with an argument `template` that gets passed to the `y` arg of `getTileExtents()`.

``` r
library(terra)
#> terra 1.7.78
```

``` r
x <- rast(system.file("ex/elev.tif", package = "terra"))
getTileExtents(x, y = fileBlocksize(x))
#>          xmin     xmax     ymin     ymax
#> [1,] 5.741667 6.533333 49.83333 50.19167
#> [2,] 5.741667 6.533333 49.47500 49.83333
#> [3,] 5.741667 6.533333 49.44167 49.47500
```

<sup>Created on 2024-07-17 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

njtierney:
I like the idea of a template object being passed along.

I think that if we have default behaviour for not specifying nrow/ncol we would be introducing multiple APIs in the same function `tar_terra_tiles`:

1. Specify ncol/nrow
2. Don't specify ncol/nrow, use blocksize by default

This also introduces additional complexity into the code now to manage these two approaches. It's not a lot of extra complexity, but I would prefer to avoid scope creep early.

As a user I think this could be harder to understand / predict what `rast_2x2`, and `rast_blocksize` are doing:

```r
  tar_target(
    elev_file,
    system.file("ex/elev.tif", package="terra"),
    format = "file"
  ),
  tar_terra_rast(
    elev_rast,
    terra::rast(elev_file)
  ),
  tar_terra_tiles(
    name = rast_2x2,
    raster = elev_rast,
    ncol = 2,
    nrow = 2
  ),
  tar_terra_tiles(
    name = rast_blocksize,
[...]

PR #84, opened 2024-07-19
Tiling improvments

Aariq:
This follows the discussion in #83 and implements the version where `tar_terra_tiles()` takes a function name or anonymous function.

I like this because:
- You don't have to think too much about creating a list object
- It minimizes the number of targets the user has to create

Limitations:
- you *cannot* create a tile specification "manually" and supply it—you'd have to write your own helper function instead.

Possible tweaks:
- it might be possible to instead pass an *expression*.  E.g. instead of `tile_blocklist` it would be `tile_blocklist(my_raster)` and instead of `\(x) tile_n(x, 2, 2)` it would be `tile_n(my_raster, 2, 2)` where `my_raster` would be the same `SpatRaster` object/target supplied to the `raster` argument of `tar_terra_tiles()`.
- With that tweak, you could probably construct the extent object as a separate target and just pass the target name as that expression instead of a function.
- You could more easily screw this up though by accidentally using a different raster for your tile template.

Let me know what you think.  If you like the idea, I can properly deprecate `creat_tile_exts()` which in this PR is renamed to `tile_n()`.


R/tar_terra_tiles.R
```diff
@@ -233,30 +234,34 @@ set_window <- function(raster, window) {
...
 #'
-#' A wrapper around [terra::getTileExtents()] for creating a tile specification.
-#' Rather than having to supply a template, one is created from the original
-#' raster extent and CRS using `ncol` and `nrow`.  The output is a list of named
-#' numeric vectors that can be coerced to SpatExtent objects with
-#' [terra::ext()].
+#' `tile_blocksize()` creates extents using the raster's native blocksize (see
+#' [terra::fileBlocksize()]), which should be more memory efficient. `tile_n()`
+#' allows specification of a number of rows and columns to split the raster
+#' into.  E.g. nrow = 2 and ncol = 2 would create 4 tiles.
```
njtierney, line 247:
```suggestion
#' into.  E.g. nrow = 2 and ncol = 2 would create 4 tiles (because it specifies a 2x2 matrix, which has 4 elements).
```

njtierney (changes requested):
What are your thoughts on `tile_n` becoming `tile_grid` or `tile_spec`. To me, the `n` implies that I would suggest a total number of elements - similar to how in dplyr there is `n_groups`, `n_distinct`. Whereas in specifying a grid or a spec, we are suggesting that there is more information to be specified.

A `tile_n` function could then be specified that must be a square number, and that could then specify breaking it into `n` tiles. e.g., `tile_n(4)` would be a 2x2.

I am happy to write this out if you'd like.

Aariq:
> What are your thoughts on `tile_n` becoming `tile_grid` or `tile_spec`. To me, the `n` implies that I would suggest a total number of elements - similar to how in dplyr there is `n_groups`, `n_distinct`. Whereas in specifying a grid or a spec, we are suggesting that there is more information to be specified.
> 
> A `tile_n` function could then be specified that must be a square number, and that could then specify breaking it into `n` tiles. e.g., `tile_n(4)` would be a 2x2.
> 
> I am happy to write this out if you'd like.

`tile_grid()` makes sense to me.  I think that should be sufficient and we probably don't need `tile_n()`

njtierney (approved):
Looks good to me!

njtierney:
> `tile_grid()` makes sense to me. I think that should be sufficient and we probably don't need `tile_n()`

Great! I will open a separate issue to log the idea of `tile_n()` in case we think it might be useful later on.

PR #85, opened 2024-07-22
Hard code `iteration = "list"`

### LINKED ISSUE #81: only allow `iteration = "list"` for terra targets
Only `iteration = "list"` seems to work for `tar_terra_*()`, so perhaps that should just be hard-coded.

Aariq:
Closes #81 by hard-coding `iteration = "list"` for all `tar_` functions.

Feedback requested:

- Where to set `iteration = "list"`---in the defaults for `tar_stars()`, the call to `tar_stars_raw()`, the defaults for `tar_stars_raw()`, or in the call to `targets::tar_target_raw()`?
- Should I add a note somewhere in documentation explaining *why* `iteration` has to be `"list"`?


njtierney (approved):
Looks good to me!

njtierney:
RE:

> Feedback requested:
> 
> * Where to set `iteration = "list"`---in the defaults for `tar_stars()`, the call to `tar_stars_raw()`, the defaults for `tar_stars_raw()`, or in the call to `targets::tar_target_raw()`?

So I think it would make sense to have this hard coded as the default - so put it inside of `tar_stars_raw()`.

> * Should I add a note somewhere in documentation explaining _why_ `iteration` has to be `"list"`?

This would be great - either as a `#' @note` or `#' @description`. We can also log this as an issue to revisit down the track?

Issue #89, opened 2024-08-21
`tar_terra_rast()` renames layers


Aariq:
Not sure how we missed this, but `tar_terra_rast()` seems to re-name raster layers based on the target name.

``` r
targets::tar_script({
    library(targets)
    library(geotargets)
    library(terra)
    list(
        tar_target(
            my_file,
            system.file("ex/logo.tif", package = "terra"),
            format = "file"
        ),
        tar_terra_rast(
            my_map,
            terra::rast(my_file)
        )
    )
})
targets::tar_make()
#> terra 1.7.78
#> ▶ dispatched target my_file
#> ● completed target my_file [0.003 seconds]
#> ▶ dispatched target my_map
#> ● completed target my_map [0.006 seconds]
#> ▶ ended pipeline [0.156 seconds]
targets::tar_load_everything()

library(terra)
#> terra 1.7.78
names(my_map)
#> [1] "my_map_1" "my_map_2" "my_map_3"
names(rast(my_file))
#> [1] "red"   "green" "blue"
```

<sup>Created on 2024-08-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Aariq:
This appears to apply to `tar_terra_tiles()` as well.  A `tar_terra_tiles()` target called `rast_split` gives a list of targets with the layer names that look like `rast_split~04ea3060_1, rast_split~04ea3060_2, rast_split~04ea3060_3`

Aariq:
This is related to #58, but layer names should be able to be stored in the tif itself rather than in an aux.xml file.  E.g. there is no "ex/logo.tif.aux.xml" file in the `terra` package:

```
> system.file("ex/logo.tif.aux.xml", package = "terra")
[1] ""
```

Aariq:
Ah, nevermind—I guess I misunderstood the difference between bands and layers.  logo.tif has _bands_ for RGB and these get written to a PAM auxilliary file, but if you do `rast("logo.tif") + 1` it converts those bands to layers for some reason (???) and then those layer names aren't lost.  Going to modify the test in #90 add a note to #58, and close this.

PR #90, opened 2024-08-21
Fix `tile_blocksize()` to work with multi-layer rasters

### LINKED ISSUE #87: tile_blocksize() may fail with multi-layer rasters
https://github.com/njtierney/geotargets/blob/ea1f9fe737ebcc64a9e6bd88f589e7554e87f17f/R/tar_terra_tiles.R#L289C5-L289C76

Not sure how I missed this, but `getTileExtents()` expects 1 or 2 numbers and `fileBlocksize()` appears to return two number *per layer* as a matrix.  Need to add a test to confirm and then fix this.

Would the blocksize ever differ among layers of the same raster?  Seems like they shouldn't, since they are required to have the same resolution and projection.  If that is true, I'll just take the first row.

### LINKED ISSUE #88: Add multiplicative factor arg to `tile_blocksize()`
Let's say tiles are very small and you'd rather save out tiles that are 10 blocks by 10 blocks than 1 block by 1 block.  Would be nice to do something like `tile_blocksize(raster, mult = 10)`.  Thoughts on what to name this argument?  `n_blocks`? `mult`?

All you'd need is something like this:

``` r
tile_blocksize <- function(raster, mult = 1) {
    tile_ext <- terra::getTileExtents(raster, terra::fileBlocksize(raster) * mult)
    n_tiles <- seq_len(nrow(tile_ext))
    tile_list <- lapply(
        n_tiles,
        \(i) tile_ext[i,]
    )
    tile_list
}
```

Aariq:
Closes #87 by using only the first row of output from `terra::fileBlocksize()`, which outputs a number of rows and columns *for each layer*.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,10 +1,10 @@
 # geotargets (development version)
 
 * Created `tar_stars()` and `tar_stars_proxy()` that create `stars` and `stars_proxy` objects, respectively.
-* Created `tar_terra_tiles()`, a "target factory" for splitting a raster into multiple tiles with dynamic branching.
-* Removes the `iteration` argument from all `tar_*()` functions.  `iteration` now hard-coded as `"list"` since it is the only option that works (for now at least).
-* Created two helper functions for use in `tar_terra_tiles()`: `tile_grid()` and `tile_blocksize()`.
+* Created `tar_terra_tiles()`, a "target factory" for splitting a raster into multiple tiles with dynamic branching (#69).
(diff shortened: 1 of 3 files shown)


njtierney:
This looks good to me - can you add a NEWS bullet to link to issues  #87 and #89 ?

Aariq:
I added some issue numbers to NEWS, but I'll point out that this is still the version where `tile_blocksize()` was first introduced, so from that point of view it's not really a bug fix, but just actually properly implementing things.

njtierney:
Yup, totally agree this isn't really a bug fix per se. Thanks for adding those! I'm happy for you to merge this :)

PR #91, opened 2024-09-03
Add package logo

Aariq:
Adds our wonderful new package logo from Hubert.


Aariq:
Added the original files in a `logo/` folder that gets `.Rbuildignore`d

njtierney (approved):
Looks good to me! Thanks so much, Eric! :)

PR #96, opened 2024-09-11
Add `tile_n()`

### LINKED ISSUE #86: `tile_n()` - as part of `tar_terra_tiling`
A `tile_n` function could be specified that must be a square number, and that could then specify breaking it into n tiles. e.g., tile_n(4) would be a 2x2.

This could be useful if you know that you want to split up something into 9 pieces because you know you've got 9 cores/cpus/workers? But it might not be necessary.

Related to #83 and #84

Aariq:
Closes #86 by adding a `tile_n()` helper.  The only open question is what to do when `n` is odd?  Keep it as rectangular as possible and round up the number of tiles or just do `n` rows and 1 column?  Not sure if one has a performance advantage over the other.


R/tar_terra_tiles.R
```diff
@@ -295,3 +322,44 @@ tile_blocksize <- function(raster) {
...
+#' @rdname tile_helpers
+tile_n <- function(raster, n) {
+    if (!rlang::is_integerish(n)) {
+        rlang::abort("`n` must be an integer.")
+    }
+    sq <- sqrt(n)
+    sq_round <- floor(sq)
+    quotient <- n/sq_round
+
+    if (rlang::is_integerish(quotient)) { #if even
```
njtierney, line 335:
```suggestion
    is_even <- rlang::is_integerish(quotient)
    is_odd <- !is_even
    if (is_even) {
```

R/tar_terra_tiles.R
```diff
@@ -295,3 +322,44 @@ tile_blocksize <- function(raster) {
...
+        nrow <- sq_round
+        ncol <- n/nrow
+    } else { #if odd
```
njtierney, line 338:
```suggestion
    }
    if (is_odd) {
```

njtierney (commented):
Just a couple of small ideas, otherwise looks good!

PR #97, opened 2024-09-12
Adds GDAL min version requirement for writing .shz

### LINKED ISSUE #71: Require minimum version of GDAL?
Certain features (e.g. SOZip shapefile use by `tar_terra_vect()` with filetype = "ESRI Shapefile") require GDAL >= 3.7.  However, that release was only about a year ago.  Running tests on GDAL 3.0.4 gives the following:

```
Error (test-tar-terra.R:62:5): tar_terra_vect() works
<tar_condition_run/tar_condition_targets/rlang_error/error/condition>
Error: Error running targets::tar_make()
Last error message:
    the write() function in tar_format() must not create a directory. Found directories inside the data store where there should only be files: _targets/objects/test_terra_vect_shz

```
I think there are a few options:
1. Require GDAL >= 3.7
2. Check for GDAL >= 3.7 and error if older
3. Check GDAL version and fallback to a different system when version requirement isn't met (e.g. using `utils::zip()` instead of the .shz extension)

Aariq:
Closes #71 by adding a check for GDAL 3.1 or greater.  Not sure if there is a way to write a test for this

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -15,15 +15,13 @@ check_pkg_installed <- function(pkg, call = rlang::caller_env()) {
   }
 }
 
-check_gdal_version <- function(min_version = "3.1") {
-    terra_gdal <- numeric_version(terra::gdal(lib = "gdal"))
-    version <- numeric_version(min_version)
-    terra_gdal < min_version
+gdal_version <- function() {
+    numeric_version(terra::gdal(lib = "gdal"))
 }
 
 check_gdal_shz <- function(min_version = "3.1",
                            call = rlang::caller_env()){
-    if (check_gdal_version(min_version)) {
+    if (gdal_version() < numeric_version(min_version)) {
         cli::cli_abort(
             message = c(
                 "Must have {.pkg GDAL} version {.val {min_version}} or greater",
--- a/R/utils.R
+++ b/R/utils.R
@@ -15,6 +15,26 @@ check_pkg_installed <- function(pkg, call = rlang::caller_env()) {
   }
 }
 
+check_gdal_version <- function(min_version = "3.1") {
+    terra_gdal <- numeric_version(terra::gdal(lib = "gdal"))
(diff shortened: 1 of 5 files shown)


njtierney:
@Aariq I can make these changes to the tests if you like?

Aariq:
> @Aariq I can make these changes to the tests if you like?

Sure, if you've got an idea of how to mock different GDAL versions, go for it!

njtierney:
I think this should work! I got a bunch of failing tests locally but I'm curious to see how this goes on CI here. Happy for you to merge if you like it :)

tests/testthat/test-check-gdal-shz.R
```diff
@@ -0,0 +1,9 @@
+test_that("GDAL version checked checked", {
+    local_mocked_bindings(
+        check_gdal_version = function(...) TRUE
```
Aariq, line 3: What about going _deeper_ and using `local_mocked_bindings()` to get `terra::gdal(lib = "gdal")` to return `3.0.0`?

R/utils.R
```diff
@@ -15,6 +15,26 @@ check_pkg_installed <- function(pkg, call = rlang::caller_env()) {
   }
 }
 
+check_gdal_version <- function(min_version = "3.1") {
+    terra_gdal <- numeric_version(terra::gdal(lib = "gdal"))
+    version <- numeric_version(min_version)
+    terra_gdal < min_version
```
Aariq, line 21: Also feels like the output of this is reversed.  It makes more sense to me if it would return FALSE when the min version is not met.  I.e. `terra_gdal >= min_version`

Aariq:
Ended up going with a simpler wrapper that does less, `gdal_version()`, instead of `check_gdal_version()`.  DId not know about `local_mocked_bindings()` before!  Thanks for that!

Issue #101, opened 2024-10-02
New, easier way to pass GDAL format options to `tar_format()` functions


Aariq:
The latest release of `targets` gives us a [`substitute` argument](https://docs.ropensci.org/targets/reference/tar_format.html?q=tar_format#arg-substitute) to `tar_format()` (which looks a lot cleaner and simpler than using [tar_resources_custom_format()](https://docs.ropensci.org/targets/reference/tar_resources_custom_format.html) to pass things through to the write function as environment variables.  Might be worth exploring for use in `geotargets`.

PR #102, opened 2024-10-02
Simpler `tar_format()` with `targets` 1.8.0

Aariq:
The newest release of `targets` (1.8.0) introduces a new way to pass arguments to custom read and write functions inside of `tar_format()` using a `substitute` argument (#101).  This makes the code a little bit cleaner (no more passing of environment variables with the `resources` argument) and happens to fix an issue where changing the `filetype` argument didn't invalidate the target (#103).

Since we use @inheritParams quite a bit, the documentation for `geotargets` is also updated, sometimes in ways that seem less than ideal (e.g. mentioning specifically `tar_target()` and `tar_target_raw()`.  I suggest that any edits to documentation be dealt with in a separate PR.

### DIFF
--- a/R/tar-stars.R
+++ b/R/tar-stars.R
@@ -213,7 +213,7 @@ tar_stars_raw <- function(name,
         format = targets::tar_format(
             read = function(path) {
                 if (ncdf) {
-                    check_pkg_installed("ncmeta")
+                    geotargets:::check_pkg_installed("ncmeta")
                     stars::read_ncdf(path, proxy = proxy)
                 } else if (isTRUE(mdim)) {
                     stars::read_mdim(path, proxy = proxy)
--- a/R/tar-stars.R
+++ b/R/tar-stars.R
@@ -204,52 +204,6 @@ tar_stars_raw <- function(name,
 
     driver <- rlang::arg_match0(driver, drv$name)
 
-    .read_stars <- eval(substitute(
-        function(path) {
-
-        ## TODO: it appears envvar custom resources do not work in read function?
-        READ_FUN <- stars::read_stars
-        #  mdim <- as.logical(Sys.getenv("GEOTARGETS_GDAL_RASTER_MDIM", unset = FALSE))
-        if (mdim) {
-            READ_FUN <- stars::read_mdim
-        }
-
-        # ncdf <- as.logical(Sys.getenv("GEOTARGETS_USE_NCMETA", unset = FALSE))
-        if (ncdf && requireNamespace("ncmeta")) {
-            READ_FUN <- stars::read_ncdf
-        }
-
-        # proxy <- as.logical(Sys.getenv("GEOTARGETS_PROXY", unset = FALSE))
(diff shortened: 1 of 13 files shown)


Aariq:
Huh, I think this fixes #103, probably because of what the format function looks like when the pipeline is evaluated.  Because it fixes #103 (based on informal prelim tests), I'm going to go ahead and do this for all the functions.

R/tar-terra-vect.R
```diff
@@ -91,93 +92,47 @@ tar_terra_vect <- function(name,
...
+            write = function(object, path) {
+                terra::writeVector(
+                    object,
+                    filename = ifelse(filetype == "ESRI Shapefile", paste0(path, ".shz"), path),
+                    filetype = filetype,
+                    overwrite = TRUE,
+                    options = gdal
+                )
+                if (filetype == "ESRI Shapefile") {
+                    file.rename(paste0(path, ".shz"), path)
```
njtierney, line 118:
```suggestion
                    file.rename(file.path(path, ".shz"), path)
```

Perhaps not needed but it won't hurt

tests/testthat/test-tar-terra.R
```diff
@@ -166,3 +166,38 @@ targets::tar_test("user resources are passed correctly", {
...
+                r,
+                rast(system.file("ex/elev.tif", package="terra")),
+                filetype = "GTiff"
+            )
+        )
+    })
+    tar_make()
+    second <- tar_meta()$time[2]
+    #There's gotta be a better way to test if a target is invalidated
+    expect_gt(second, first)
```
njtierney, line 201: I'm not sure if there is a good way, but I was really curious how you were going to do this and I like what you've done!

njtierney (changes requested):
A couple of really minor things, but overall this looks good to me!

njtierney:
Thanks for doing this, this cleans up a lot of code, it looks really nice! :rocket:

R/tar-terra-vect.R
Aariq: Unfortunately that's not going to work:
```
> file.path("path/to/file", ".shz")
[1] "path/to/file/.shz"
```
`fs::path_ext_set` would work, but not worth adding a dependency

R/tar-stars.R
```diff
@@ -204,63 +204,49 @@ tar_stars_raw <- function(name,
...
         pattern = pattern,
         packages = packages,
         library = library,
         format = targets::tar_format(
-            read = .read_stars,
-            write = .write_stars,
-            marshal = function(object) object, # Not currently used
-            unmarshal = function(object) object
+            read = function(path) {
+                if (ncdf && requireNamespace("ncmeta")) {
```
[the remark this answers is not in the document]
Aariq: I did this, but unfortunately because it's inside of `tar_format()` I have to use `geotargets:::check_pkg_installed()` which gives a 'Note' on check.  I think that's fine though and we can explain it to CRAN

Issue #92, opened 2024-09-11
Silence warning from `tar_terra_sprc()`


Aariq:
Using `tar_terra_sprc()` results in the warning `[rast] skipped sub-datasets (see 'describe(sds=TRUE)')`.  I'm fairly certain this is because `terra::writeRaster()` returns a SpatRaster object (rather than a file path, which is what I'd expect) and it only returns the first file in the sprc when other rasters are added with `APPEND_SUBDATASET=YES`.  Unless there is a way to change the behavior of `writeRaster()` to not return the result of `rast()`, it might be better to just silence this warning in the write function of `tar_terra_sprc()`

PR #104, opened 2024-10-09
Suppress `tar_terra_sprc()` warning

Aariq:
Addresses #92 by wrapping the write function in `withCallingHandlers()` to capture the uninformative warning `[rast] skipped sub-datasets ...`.  The warning originates from `terra::writeRaster()` when it is used to write a subdataset because the return value of `writeRaster()` is `rast(<file>)` and that file can't be correctly read by `rast()` if it has subdatasets with different projections.

This was difficult to add a test for since warnings from write functions apparently don't bubble up to the output of `tar_meta()`.  I added a test for a version of the write function instead.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,7 @@
 * Created utility function `set_window()` mostly for internal use within `tar_terra_tiles()`.
 * Removes the `iteration` argument from all `tar_*()` functions.  `iteration` now hard-coded as `"list"` since it is the only option that works (for now at least).
 * Added the `description` argument to all `tar_*()` functions which is passed to `tar_target()`.
-* Suppressed the warning "[rast] skipped sub-datasets" from `tar_terra_sprc()`, which is misleading in this context (#92, #104)
+* Suppressed the warning "[rast] skipped sub-datasets" from `tar_terra_sprc()`, which is misleading in this context (#92, #104).
 * Requires GDAL 3.1 or greater to use "ESRI Shapefile" driver in `tar_terra_vect()` (#71, #97)
 
 # geotargets 0.1.0 (14 May 2024)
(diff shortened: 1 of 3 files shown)


tests/testthat/test-tar-terra-sprc.R
```diff
@@ -1,60 +1,60 @@
...
+    )
+    targets::tar_script({
+        elev_scale <- function(z = 1) {
+            terra::rast(
+                system.file(
+                    "ex",
+                    "elev.tif",
+                    package = "terra"
+                )
+            ) * z
```
Aariq, line 51: Sorry, this was all just fixing indenting--there's no real changes here

Aariq:
Warnings from `tar_format()` now show up in `tar_meta()` in the dev version of `targets`.  Can simplify this test eventually. https://github.com/ropensci/targets/discussions/1345#discussioncomment-10908585

NEWS.md
```diff
@@ -6,6 +6,7 @@
 * Created utility function `set_window()` mostly for internal use within `tar_terra_tiles()`.
 * Removes the `iteration` argument from all `tar_*()` functions.  `iteration` now hard-coded as `"list"` since it is the only option that works (for now at least).
 * Added the `description` argument to all `tar_*()` functions which is passed to `tar_target()`.
+* Suppressed the warning "[rast] skipped sub-datasets" from `tar_terra_sprc()`, which is misleading in this context (#92, #104)
```
njtierney, line 9:
```suggestion
* Suppressed the warning "[rast] skipped sub-datasets" from `tar_terra_sprc()`, which is misleading in this context (#92, #104).
```

njtierney (approved):
Looks good to me! I haven't used calling handlers before or https://rlang.r-lib.org/reference/cnd_muffle.html before, so that is cool to see!

PR #62, opened 2024-04-29
[draft] Implement SOZip storage of terra targets

brownag:
This is a draft PR to implement an option (`zipfile=TRUE`) to write/read Seek-Optimized ZIP (SOZIP) files in the target store for #37. This works for tar_terra* SpatVector and SpatRaster methods,

Rather than attempting to create (non-SOZIP) ZIP files independently using `utils::zip()` or similar this PR uses one of two (simpler, but somewhat more limited) pathways available thru existing GDAL drivers.

 - Specific drivers that support direct write of SOZIP via a file extension for SOZIP (ESRI Shapefile and Geopackage, as of GDAL 3.7) use that path for writing. So, ESRI Shapefile uses .shz extension (as it already does), and GeoPackage uses .gpkg.zip. Read is done via /vsizip/...

 -  All other drivers use `"/vsizip/{path/to/zipfile/target}/target"` generic data source path for write and read. This is not supported by all drivers currently, but works for things like Parquet and GeoTIFF. GeoTIFF requires specific GDAL options (STREAMABLE_OUTPUT=YES, COMPRESS=NONE). 

Examples of the above cases have been added to tests.

Raster:
``` r
targets::tar_script({
    list(
        geotargets::tar_terra_rast(
            test_terra_rast2,
            terra::rast(system.file("ex/elev.tif", package = "terra")),
            gdal = c("STREAMABLE_OUTPUT=YES", "COMPRESS=NONE"),
            zipfile = TRUE
        ),
        geotargets::tar_terra_rast(
            test_terra_rast3,
            terra::rast(system.file("ex/elev.tif", package = "terra")),
            filetype = "GPKG",
            zipfile = TRUE
        )
    )
})
targets::tar_make()
#> ▶ dispatched target test_terra_rast2
#> ● completed target test_terra_rast2 [0.007 seconds]
#> ▶ dispatched target test_terra_rast3
#> ● completed target test_terra_rast3 [0.004 seconds]
#> ▶ ended pipeline [0.202 seconds]
#> Warning messages:
#> 1: /vsizip/{_targets/scratch/test_terra_rast2}/test_terra_rast2, band 1: Cannot modify metadata at that point in a streamed output file (GDAL error 6) 
#> 2: /vsizip/{_targets/scratch/test_terra_rast2}/test_terra_rast2, band 1: Cannot modify metadata at that point in a streamed output file (GDAL error 6) 
[...]

PR #106, opened 2024-10-22
Use _geotargets cache for files, _targets for `PackedSpatRaster`s

Aariq:
A slightly different take on #63 that uses an option (currently called `use_cache`) to switch on behavior to deal with #58 instead of adding an additional `tar_terra_*()` function.

When `use_cache` is `TRUE`, the raster is written with `terra::wrapCache()` to `cache_dir` (default `_geotargets/`) and the output (a `PackedSpatRaster` object) is saved to the targets store with `saveRDS()`.  This allows 'sidecar' files to exist and be read back in by the read function (`terra::unwrap(readRDS(path))`).

I'd love any feedback on **naming things** and **where to document this**.  E.g. I'm not sure that "cache" is really a good/complete descriptor for what's going on here.  `use_external_cache`? `use_raster_cache`? `use_proxy_cache`?

Things to do in other PRs before this one is ready:
- [ ] update parameter documentation to customize some currently inherited documentation now that help files for `tar_target()` and `tar_target_raw()` have been merged in `targets` (#107, #108)

Things to do in this PR before it is ready:
- [ ] `document()`
- [ ] Make sure management of the `_geotargets/` cache is user-friendly.  E.g. when that cache is destroyed (by a `geotargets` function), the corresponding targets are also invalidated.
- [ ] *possibly* implement something like `tar_prune()` but for the `_geotargets/` cache.
- [ ] testing and/or documentation that this probably doesn't work with cloud storage


Aariq:
Hmm... now questioning whether this is better than just using zip or tar files (optionally) like in #62. Might need to do some benchmarking to see how much overhead each method adds.  If they are similar, then maybe zip/tar files (with low or no compression) are better since it doesn't involve managing a separate cache.

PR #108, opened 2024-10-22
custom documentation for name, command, and pattern params

### LINKED ISSUE #107: Update parameter documentation
Now that the documentation for `tar_target()` and `tar_target_raw()` as been combined in `targets` (https://github.com/ropensci/targets/releases/tag/1.8.0) a few of the `@params` inherited by our `tar_*()` functions don't make as much sense since they specifically mention `tar_target()` and `tar_target_raw()`.  We'll need to set our own documentation for these, perhaps just borrowing from a previous version of `targets`.

Aariq:
This closes #107 by adding documentation for the name, command, and pattern parameters to all `tar_*()` functions rather than relying on importing the documentation for these from `targets`

### DIFF
--- a/R/geotargets-option.R
+++ b/R/geotargets-option.R
@@ -5,13 +5,13 @@
 #'
 #' @param gdal_raster_driver character, length 1; set the driver used for raster
 #'   data in target store (default: `"GTiff"`). Options for driver names can be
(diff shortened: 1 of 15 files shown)


njtierney (approved):
Thanks for doing this! I've added some missing full-stops at the ends of sentences that were missing them, happy for you to merge :)

PR #109, opened 2024-10-28
Preserve metadata with zip archives

### LINKED ISSUE #58: `tar_terra_rast()` doesn't save units or varnames

Aariq:
This closes #58 and supersedes #63 and #106.

Adds a `preserve_metadata` argument to `tar_terra_rast()` that when `TRUE` switches the write and read functions to write to a tempdir and zip files before moving them to `path` and to unzip that archive before reading in with `terra::rast()`.  This is a blanket solution (I think) that should work for any raster driver.  There are notes as to why `vsizip/` wasn't used in `notes/`

Question for reviewers:
Should `terra`, `zip`, and `withr` still be in Suggests?  Now that it's more clear the main utility of `geotargets` is to work with `terra`, we might consider moving it (and the dependencies needed for `preserve_metadata`) into `Imports`


.Rbuildignore
njtierney, line 1: I like the notes folder! What do you think about put it inside `inst/notes` instead? That way we can have a little home for future things, e.g., https://github.com/tidyverse/dplyr/tree/main/inst.

notes/preserve_metadata_benchmarking.Qmd
njtierney, line 1: See above note on R buildignor

R/tar-terra-rast.R
```diff
@@ -38,6 +44,7 @@ tar_terra_rast <- function(name,
                            pattern = NULL,
                            filetype = geotargets_option_get("gdal.raster.driver"),
                            gdal = geotargets_option_get("gdal.raster.creation.options"),
+                           preserve_metadata = FALSE, #TODO: add a geotargets option for this
```
njtierney, line 47: I think options are a good idea, we should add an issue for this so we can come back to it later

R/tar-terra-rast.R
```diff
@@ -6,6 +6,12 @@
 #'   to [terra::writeRaster()]
 #' @param gdal character. GDAL driver specific datasource creation options
 #'   passed to [terra::writeRaster()]
+#' @param preserve_metadata logical. When `FALSE` (default), any auxiliary files
```
njtierney, line 9: Ever since reading https://design.tidyverse.org/boolean-strategies.html (and to some extent, https://design.tidyverse.org/argument-clutter.html ) I now always take a moment to consider TRUE/FALSE arguments as a potential opportunity to use a different set of arguments. E.g., `preserve_metadata` could instead be "drop" or "zip", perhaps? This also helps us keep open other options for dealing with this in the future. 

It does create more work to set up different options, but I think it might be worthwhile, but there are tradeoffs as well. What are your thoughts?

njtierney (changes requested):
This looks  great! As mentioned in the review, I think it might be useful/worthwhile to consider using enum / character option instead of TRUE/FALSE for `preserve_metadata`, as discussed in https://design.tidyverse.org/boolean-strategies.html. 

Re

> Question for reviewers:
Should terra, zip, and withr still be in Suggests? Now that it's more clear the main utility of geotargets is to work with terra, we might consider moving it (and the dependencies needed for preserve_metadata) into Imports

I agree with you at `terra, `zip`, and `withr` should be in Imports now, especially since it is more or less the main goal of `geotargets` to work with `terra` - which we didn't know for sure would be the case when we started the project.

R/tar-terra-rast.R
brownag: I also think it might be worthwhile to allow for different options and use a set of character method names rather than boolean. I can imagine, for instance, future terra or GDAL versions that have better support for /vsizip/ read and write  that users might want to be able to opt-in to.

.Rbuildignore
Aariq: Ah of course, this is exactly where it belongs

R/tar-terra-rast.R
Aariq: Sounds great, but I don't think we can do this since the SpatRaster doesn't yet exist when the target object (including `tar_format()` with read/write functions) is created

Aariq: > I think options are a good idea, we should add an issue for this so we can come back to it later

I can implement it in this PR—I just forgot (exactly why I added a #TODO!)

R/tar-terra-rast.R
Aariq: Makes sense.  I'll change the options to `"drop"` and `"zip"` and update the documentation accordingly.  Might be able to add `"tar"` (you can make tar archives without compression, but I can't get base R `tar()` to work) and `"vsizip"` as options in the future.

notes/preserve_metadata_benchmarking.Qmd
```diff
@@ -0,0 +1,230 @@
...
+```{r}
+#"pretend" _targets/ store
+target_store <- "_test/object"
+dir_create(target_store)
+```
+
+## Using an archive
+
+Getting the archive to not contain a bunch of junk paths is tricky (impossible?) with base R's `zip()` or `tar()` functions.
+Plus, they have potentially different behaviors on different operating systems.
```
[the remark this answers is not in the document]
Aariq: I think the one potential benefit of using base R's `tar()` is that you can create archives without compression, which I assume is faster (but I could be wrong)

Aariq:
Ok, I've got the geotargets option for preserve metadata worked out, if anyone wants to give feedback on naming things.

R/tar-terra-rast.R
```diff
@@ -118,3 +120,53 @@ tar_terra_rast <- function(name,
...
+    if (preserve_metadata == "zip") {
+        function(path) {
+            tmp <- withr::local_tempdir()
+            zip::unzip(zipfile = path, exdir = tmp)
+            terra::rast(file.path(tmp, basename(path)))
+        }
+    } else if (preserve_metadata == "drop") {
+        function(path) terra::rast(path)
+    }
+}
```
njtierney, line 134: Could these be potentially rewritten as switch statements? E.g.,

```r
switch(
  preserve_metadata,
  zip = function(path) {
    tmp <- withr::local_tempdir()
    zip::unzip(zipfile = path, exdir = tmp)
    terra::rast(file.path(tmp, basename(path)))
  },
  drop = function(path) terra::rast(path)
)
```

njtierney (commented):
Looks awesome!

Just some small changes! Not sure if the switch statement is necessary, but it might be useful, maybe?

njtierney (approved):
Looks good to me, thanks so much, @Aariq !

PR #111, opened 2024-10-31
Vignette for dynamic branching with tiles

Aariq:
Adds a vignette for dynamic branching with `tar_terra_tiles()`.  Originally, I thought I'd use this to show the benefits of using `tile_blocksize()` over creating arbitrarily sized tiles, but that doesn't seem to be the case.  I couldn't come up with an example where it was actually faster to use `tile_blocksize()` 🤷‍♂️.  So instead this just shows how to create a dynamic-branching-over-tiles workflow using only `tile_blocksize()`.


vignettes/tar_terra_tiles.Rmd
```diff
@@ -0,0 +1,386 @@
...
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>",
+  fig.align = "center",
+  fig.height = 5,
+  fig.width = 5
+)
+if (identical(Sys.getenv("NOT_CRAN", unset = "false"), "false")) {
+  knitr::opts_chunk$set(eval = FALSE)
+}
```
Aariq, line 25: @njtierney, I just copied this from the other vignette you started—can you explain this?  It's the opposite of what I'd expect—eval gets set to FALSE when I build locally.

Aariq: I'm going to negate it for now, but let me know if I'm missing something

Aariq:
Thanks for the additions!  It is much improved!

vignettes/tar_terra_tiles.Rmd
Aariq: Oh, I see now that `check()` sets `NOT_CRAN=true`, but maybe `build_vignettes()` doesn't? Anyway, this is back to how it was and should work now

PR #113, opened 2024-11-13
Force read into memory

Aariq:
This is a necessary addition to `tar_terra_rast()` for `SpatRaster` targets to work with cloud storage. (see more discussion in https://github.com/ropensci/targets/discussions/1367).  I added a parameter and some documentation, but haven't figured out how to "hook it up" efficiently.  When `force_memory` is true, it needs to modify the body of the read function to add ` + 0` or `* 1` to the end.  Tried messing around with `body(fun)` and `rlang::expr()` but didn't get anything satisfactory.  I'm open to ideas and alternatives! 

One alternative is to use the new content addressable storage tools in `targets` to have a custom download function that doesn't actually download the target, but rather returns a `/vsis3/` filepath to the read function.  This is not something we'd implement in `geotargets`, but would be something cool to document in a vignette.  I may play around with this in the next few weeks.

### DIFF
--- a/R/tar-terra-rast.R
+++ b/R/tar-terra-rast.R
@@ -143,9 +143,9 @@ tar_rast_read <- function(preserve_metadata) {
         zip = function(path) {
             tmp <- withr::local_tempdir()
             zip::unzip(zipfile = path, exdir = tmp)
-            terra::rast(file.path(tmp, basename(path)))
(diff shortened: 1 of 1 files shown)


Aariq:
I actually think documenting the content addressable storage workaround is probably a better way to go, but I'll leave this open for a bit just in case y'all think there might be good reasons to add a `force_memory` argument.

Issue #99, opened 2024-09-18
`tar_terra_tiles()` errors with `tile_blocksize()` helper when raster is created as a target


Aariq:
Getting a mysterious and difficult to reproduce error when `tar_terra_tiles()` is used with the `tile_blocksize()` helper and the raster target is not reading in from a file.

Here's a reprex

``` r
library(targets)
tar_dir({
    tar_script({
        library(targets)
        library(geotargets)
        library(terra)
        list(
            tar_target(
                raster_file,
                system.file("ex/elev.tif", package="terra"),
                format = "file"
            ),
            tar_terra_rast(
                raster,
                rast(raster_file)
            ),
            tar_terra_rast(
                raster2,
                raster + 3
            ),
            tar_terra_tiles(
                tiles_blocksize,
                raster = raster2,
                tile_fun = tile_blocksize
            )
        )
    })

    tar_make()
})
#> terra 1.7.78
#> ▶ dispatched target raster_file
#> ● completed target raster_file [0.003 seconds]
#> ▶ dispatched target raster
#> ● completed target raster [0.017 seconds]
#> ▶ dispatched target raster2
#> ● completed target raster2 [0.004 seconds]
#> ▶ dispatched target tiles_blocksize_exts
#> ✖ errored target tiles_blocksize_exts
#> ✖ errored pipeline [0.268 seconds]
#> Error:
#> ! Error running targets::tar_make()
#> Error messages: targets::tar_meta(fields = error, complete_only = TRUE)
#> Debugging guide: https://books.ropensci.org/targets/debugging.html
#> How to ask for help: https://books.ropensci.org/targets/help.html
#> Last error message:
#>     [aggregate] values in argument 'fact' should be > 0
#> Last error traceback:
#>     tile_blocksize(raster2)
#>     terra::getTileExtents(raster, terra::fileBlocksize(raster)[1,      ] * c...
#>     terra::getTileExtents(raster, terra::fileBlocksize(raster)[1,      ] * c...
#>     .local(x, ...)
#>     aggregate(rast(x), y)
#>     aggregate(rast(x), y)
#>     .local(x, ...)
#>     messages(x, "aggregate")
#>     error(f, x@ptr$getError())
#>     stop("[", f, "] ", emsg, ..., call. = FALSE)
#>     .handleSimpleError(function (condition)  {     state$error <- build_mess...
#>     h(simpleError(msg, call))
```

[...]

PR #114, opened 2024-11-21
Document workaround for issues with `tile_blocksize()` (#99)

Aariq:
Add documentation to `tar_terra_tiles()` and `tile_blocksize()` explaining why and how to force upstream targets to not be in memory.

Also updates other documentation because of running `document()` with `targets` 1.9.0 installed


njtierney (approved):
Looks good to me!

PR #115, opened 2024-11-21
handle check() notes

Aariq:
Deals with two `check()` notes:

- adds `fs` to Suggests (it is used in vignette)
- moves `geotargets:::check_pkg_installed()` outside of a write function so it can just be `check_pkg_installed()`

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -45,7 +45,8 @@ Suggests:
     ncmeta,
     sf,
     stars,
-    testthat (>= 3.0.0)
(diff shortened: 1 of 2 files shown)


njtierney:
Nice! Happy with this, I tried not using `fs` and using `dir.create()` instead but I got some issues with the vignette build, so happy to keep `fs` in Suggests.

Made a small tweak to the if statement to add curly braces.

I'll just wait to see if this passes R CMD Check and then will merge :)

Issue #121, opened 2024-11-28
`tar_terra_rast()` fails when temp directory is a different partion from working directory


brownag:
I noticed that the use of `file.rename()` in `tar_terra_rast(..., preserve_metadata = "zip")` does not work when the temporary directory is on a different partition. 

When creating the target the following warning occurs: `cannot rename file '/tmp/RtmpukbTP9/filedf5a4791eea/test' to '_targets/scratch/test', reason 'Invalid cross-device link'`. Subsequent attempts to load the object from the target store will fail. This occurs when running a targets script normally, but not in reprexes or tests where the working directory target store is also on the /tmp partition.

A simple workaround I found is to use `file.copy()` and `unlink()` rather than [`file.rename()` in `tar_terra_rast()`](https://github.com/njtierney/geotargets/blob/d719f97fb21e593bc40f3eca3ba1efeffce4ce24/R/tar-terra-rast.R#L176)

Output from running interactively:
```
> library(targets)
> print(getwd())
[1] "/home/andrew/workspace/geotargets"
> tar_script({
+     library(geotargets)
+     tar_terra_rast(test, 
+                    terra::rast(system.file("ex", "elev.tif", package="terra")),
+                    preserve_metadata = "zip")
+ })
> tar_make()
▶ dispatched target test
✖ errored target test
✖ errored pipeline [0.256 seconds]
Warning message:
In file.rename(file.path(tmp, basename(path)), path) :
  cannot rename file '/tmp/RtmpukbTP9/filedf5a4791eea/test' to '_targets/scratch/test', reason 'Invalid cross-device link'
Error:
! targets::tar_make() error

── Debug target test ─────────────────────────────────────────────────────────────────────────────────────
    tar_meta(test)$error
    tar_workspace(test)

── General debugging ─────────────────────────────────────────────────────────────────────────────────────
    • tar_errored()
    • tar_meta(fields = any_of("error"), complete_only = TRUE)
    • tar_workspace()
    • tar_workspaces()

── How to ────────────────────────────────────────────────────────────────────────────────────────────────
    • Debug: https://books.ropensci.org/targets/debugging.html
    • Help: https://books.ropensci.org/targets/help.html

── Last error message ────────────────────────────────────────────────────────────────────────────────────
    _store_ missing files: _targets/objects/test

── Last error traceback ──────────────────────────────────────────────────────────────────────────────────
    No traceback available.
```

[...]

PR #122, opened 2024-11-28
Fix for #121

brownag:
This is a simple fix for the issue with #121. 

Open to alternative solutions but it would be nice for this to "just work" on all platforms without modifying default temporary directory.


njtierney:
This looks good to me, just going to check it out locally and add a NEWS bullet, thanks @brownag !

njtierney:
Thanks so much for making this PR, @brownag !

PR #123, opened 2024-11-28
tar_terra_rast: implement {gdalraster} SOZip `preserve_metadata` method

brownag:
This is a draft/example PR to adds an additional option "gdalraster_sozip" for the `preserve_metadata` argument to `tar_terra_rast()`.

In the future it would be nice if the GDAL C API for working with SOZip files was exposed through terra--which could be a future option i.e. "terra_sozip" or just "sozip".

Using this utility requires GDAL >= 3.7. I have not added any version check to this yet. 
Some additional questions I have pertain to the ability to write to ZIP with multiple threads. I have not tinkered with this on HPC/multiple worker type setups.


R/tar-terra-rast.R
```diff
@@ -145,6 +149,9 @@ tar_rast_read <- function(preserve_metadata) {
             zip::unzip(zipfile = path, exdir = tmp)
             terra::rast(file.path(tmp, basename(path)))
         },
+        gdalraster_sozip = function(path) {
+          terra::rast(paste0("/vsizip/{", path, "}/"), basename(path))
+        },
```
njtierney, line 156: Do we need to worry about file.path here, or is this syntax going to be generalisable to all OS's?

R/tar-terra-rast.R
```diff
@@ -175,6 +182,38 @@ tar_rast_write <- function(filetype, gdal, preserve_metadata) {
             #move the zip file to the expected place
             file.rename(file.path(tmp, basename(path)), path)
         },
+        gdalraster_sozip = function(object, path) {
+
+            tmp <- withr::local_tempdir()
+
+            dir.create(file.path(tmp, dirname(path)), recursive = TRUE)
```
njtierney, line 193: Suggest moving `file.path(tmp, dirname(path))` out into a separate variable, since it gets referred to a few times later on in this function

R/tar-terra-rast.R
```diff
@@ -175,6 +182,38 @@ tar_rast_write <- function(filetype, gdal, preserve_metadata) {
...
+                overwrite = TRUE,
+                sozip_enabled = "YES",
+                num_threads = 1,
+                quiet = TRUE
+            )
+            # always create sozip regardless of file size (sozip_enabled = "YES")
+            # TODO: allow user control of number of threads?
+            #       how does num_threads interact multiple workers etc.?
+
+            unlink(file.path(tmp, path), )
```
njtierney: should unlink have a trailing comma here?

njtierney (changes requested):
Thanks so much for this, @brownag ! A couple of small changes requested, and I'm also wondering if/how we should think about tests for this?

R/tar-terra-rast.R
brownag: Thanks for catching, I believe I spotted this in the merge commit!

Aariq:
What do you think about using just `preserve_metadata = "sozip"`?  Simpler to type and the "engine" behind this option can change if better tooling becomes available

R/tar-terra-rast.R
brownag: ok this is done. Sorry for some commit noise on this, made a mistake that broke tests, but I think it is resolved now

Aariq:
>However I think this doesn't actually resolve all of our metadata issues as it appears .aux.json is not read by terra when specifying the path this way (see tests added in https://github.com/njtierney/geotargets/commit/02ca360a7e1515080337a319a8ef4cfdc45afe2a following the "zip" example).

It might be good to get clarity on whether this is something that _can_ be fixed in `terra` before putting too much more energy into this.  It seems possible that it might not ever be fixed because GDAL is doing the reading of the `/vsizip/` path and `terra` (with base R functions? some other mechanism?) is reading in the aux.json file.

If this is something you can figure out how to do (add the aux.json metadata back in to the `SpatRaster` after a /vsizip/ path is read) it *might* make more sense as a contribution to `terra`.

R/tar-terra-rast.R
hansvancalster: See also #127 -  the fix(es) for that issue are maybe also required here.

R/tar-terra-rast.R
brownag: The `/vsizip/` portion of the path works in genera, I've replaced the trailing slash with a call to `file.path()` in https://github.com/ropensci/geotargets/pull/123/commits/24513985bff9fa74e391ccc3d058377353be6755

brownag:
I've fixed the merge conflicts and cleaned up a bit following review comments

The goal of this PR is to implement an alternative to using `zip::zip()` for storing multi-file targets. The alternative is built in to GDAL and accessed via `gdalraster::addFilesInZip()`. This method allows for creating seek-optimized ZIP files, and also prevents the need for extraction of the target before reading back into R when `/vsizip/...` path is used on read.

njtierney (approved):
This all looks good to me, (just needs a NEWS bullet), let me know if you'd like me to merge this? I'm happy to write the NEWS bullet if you like!

njtierney:
Ah actually @brownag how do you feel about the merge conflict here?

brownag:
> This all looks good to me, (just needs a NEWS bullet), let me know if you'd like me to merge this? I'm happy to write the NEWS bullet if you like!

Done

> Ah actually @brownag how do you feel about the merge conflict here?

Added NEWS item and fixed the resulting merge conflict. That was the only remaining issue I could see

njtierney:
I keep getting this error when I render the `geotargets.Rmd` vignette locally:

```
Error:
! Error in tar_make():
  error in evaluating the argument 'x' in selecting a method for function 'sprc': error in evaluating the argument 'x' in selecting a method for function 'project': external pointer is not valid
  See https://books.ropensci.org/targets/debugging.html
```

But I don't see this error appearing on GH actions - can others confirm if they are getting this error?

brownag:
> I keep getting this error when I render the `geotargets.Rmd` vignette locally:
> 
> ```
> Error:
> ! Error in tar_make():
>   error in evaluating the argument 'x' in selecting a method for function 'sprc': error in evaluating the argument 'x' in selecting a method for function 'project': external pointer is not valid
>   See https://books.ropensci.org/targets/debugging.html
> ```
> 
> But I don't see this error appearing on GH actions - can others confirm if they are getting this error?

@njtierney After merging master to resolve differences in NEWS.md I now see this error in GH actions (e.g. https://github.com/ropensci/geotargets/actions/runs/14468868538/job/40577532981). 

I believe the issue is the "sprc" portion of the vignette was using regular `tar_target()` on a SpatRaster object, whereas we want to use `tar_terra_rast()` https://github.com/ropensci/geotargets/pull/123/commits/fdabd3575a8ebbdffd9dcf8f98c7db655d593013. 

---

Unrelated to above, but for this PR, I just fixed one test related to the behavior of `terra::metags()` on 1.8-42 (where vector output has changed to data.frame). https://github.com/ropensci/geotargets/pull/123/commits/1ebe6d2b9cdb3d1610ab8d77855c1b66b13a87b2

brownag:
> > I keep getting this error when I render the `geotargets.Rmd` vignette locally:
> > ```
> > Error:
> > ! Error in tar_make():
> >   error in evaluating the argument 'x' in selecting a method for function 'sprc': error in evaluating the argument 'x' in selecting a method for function 'project': external pointer is not valid
> >   See https://books.ropensci.org/targets/debugging.html
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But I don't see this error appearing on GH actions - can others confirm if they are getting this error?
> 
> @njtierney After merging master to resolve differences in NEWS.md I now see this error in GH actions (e.g. https://github.com/ropensci/geotargets/actions/runs/14468868538/job/40577532981).
> 
> I believe the issue is the "sprc" portion of the vignette was using regular `tar_target()` on a SpatRaster object, whereas we want to use `tar_terra_rast()` [fdabd35](https://github.com/ropensci/geotargets/commit/fdabd3575a8ebbdffd9dcf8f98c7db655d593013).

Note that it seems we did not catch this earlier because the object `r` used to build the SpatRasterCollection was just created in the same/first pipeline run, so the pointer was still "good", and the object was not read from target store. In new version of targets (1.11.1) the error triggers whereas on older versions it did not.

Aariq:
> Note that it seems we did not catch this earlier because the object `r` used to build the SpatRasterCollection was just created in the same/first pipeline run, so the pointer was still "good", and the object was not read from target store. In new version of targets (1.11.1) the error triggers whereas on older versions it did not.

Ah, this is related to new defaults with `memory = "auto"`.  Setting memory to transient is a good way to catch this kind of stuff.

Aariq (approved):
looks good!

PR #131, opened 2024-12-19
Fix missing data when `preserve_metadata = "zip"`

### LINKED ISSUE #127: problem tar_terra_rast(preserve_metadata = "zip")
https://github.com/njtierney/geotargets/blob/3f8f1daa96d9f64c02e4574329a6f8c53763cf20/R/tar-terra-rast.R#L152-L153

The above lines cause an error in my targets pipeline:

image

I'm 99% sure the error is caused by tar_terra_rast itself. When debugging the target I could run the function I wrote to process the rasters  without a problem.

I tried making a small reprex, but unfortunately could not reproduce the error. As you can see from the error, the file.path command is trying to paste together two absolute paths, which fails.
I looked into the source code, but it was unclear to me where the `path` object is declared.

Thanks for working on this great package! Hope one of you can solve the issue or tell me what I am doing wrong.

<details>

<summary>session info</summary>

```
─ Session info 
 setting  value
 version  R version 4.4.2 (2024-10-31 ucrt)
 os       Windows 10 x64 (build 19045)
 system   x86_64, mingw32
 ui       RStudio
 language (EN)
 collate  Dutch_Belgium.utf8
 ctype    Dutch_Belgium.utf8
 tz       Europe/Brussels
 date     2024-12-17
 rstudio  2024.09.1+394 Cranberry Hibiscus (desktop)
 pandoc   3.1.12.1 @ C:\\Users\\HANS_V~1\\AppData\\Local\\Pandoc\\pandoc.exe

─ Packages 
 ! package       * version    date (UTC) lib source
   assertthat    * 0.2.1      2019-03-21 [1] CRAN (R 4.4.0)
   backports       1.5.0      2024-05-23 [1] CRAN (R 4.4.0)
   base64url       1.4        2018-05-14 [1] CRAN (R 4.4.0)
   callr           3.7.6      2024-03-25 [1] CRAN (R 4.4.0)
   cli             3.6.3      2024-06-21 [1] CRAN (R 4.4.1)
   clipr           0.8.0      2022-02-22 [1] CRAN (R 4.4.0)
   codetools       0.2-20     2024-03-31 [2] CRAN (R 4.4.2)
   crew            0.10.2     2024-11-15 [1] CRAN (R 4.4.2)
   data.table      1.16.2     2024-10-10 [1] CRAN (R 4.4.2)
   evaluate        1.0.1      2024-10-10 [1] CRAN (R 4.4.2)
   fansi           1.0.6      2023-12-08 [1] CRAN (R 4.4.0)
   fortunes        1.5-4      2016-12-29 [1] CRAN (R 4.4.0)
   geotargets    * 0.2.0.9000 2024-11-29 [1] https://njtierney.r-universe.dev (R 4.4.2)
   getip           0.1-4      2023-12-10 [1] CRAN (R 4.4.0)
   glue            1.8.0      2024-09-30 [1] CRAN (R 4.4.2)
   igraph          2.1.1      2024-10-19 [1] CRAN (R 4.4.2)
   knitr           1.49       2024-11-08 [1] CRAN (R 4.4.2)
   later           1.3.2      2023-12-06 [1] CRAN (R 4.4.0)
   lifecycle       1.0.4      2023-11-07 [1] CRAN (R 4.4.0)
   magrittr        2.0.3      2022-03-30 [1] CRAN (R 4.4.0)
   mirai           1.3.1      2024-11-15 [1] CRAN (R 4.4.2)
   nanonext        1.4.0      2024-12-02 [1] CRAN (R 4.4.2)
   pillar          1.9.0      2023-03-22 [1] CRAN (R 4.4.0)
   pkgconfig       2.0.3      2019-09-22 [1] CRAN (R 4.4.0)
   processx        3.8.4      2024-03-16 [1] CRAN (R 4.4.0)
   promises        1.3.0      2024-04-05 [1] CRAN (R 4.4.0)
   ps              1.8.1      2024-10-28 [1] CRAN (R 4.4.2)
   qs              0.27.2     2024-10-01 [1] CRAN (R 4.4.2)
   R6              2.5.1      2021-08-19 [1] CRAN (R 4.4.0)
   ragg            1.3.3      2024-09-11 [1] CRAN (R 4.4.2)
   RApiSerialize   0.1.4      2024-09-28 [1] CRAN (R 4.4.1)
   Rcpp            1.0.13-1   2024-11-02 [1] CRAN (R 4.4.2)
 D RcppParallel    5.1.9      2024-08-19 [1] CRAN (R 4.4.1)
   rlang           1.1.4      2024-06-04 [1] CRAN (R 4.4.1)
   rprojroot       2.0.4      2023-11-05 [1] CRAN (R 4.4.0)
   rstudioapi      0.17.1     2024-10-22 [1] CRAN (R 4.4.2)
   secretbase      1.0.3      2024-10-02 [1] CRAN (R 4.4.1)
   sessioninfo     1.2.2      2021-12-06 [1] CRAN (R 4.4.0)
   stringfish      0.16.0     2023-11-28 [1] CRAN (R 4.4.0)
   systemfonts     1.1.0      2024-05-15 [1] CRAN (R 4.4.0)
   targets       * 1.8.0      2024-10-02 [1] CRAN (R 4.4.2)
   terra         * 1.7-83     2024-10-14 [1] CRAN (R 4.4.2)
   textshaping     0.4.0      2024-05-24 [1] CRAN (R 4.4.0)
   tibble        * 3.2.1      2023-03-20 [1] CRAN (R 4.4.0)
   tidyselect      1.2.1      2024-03-11 [1] CRAN (R 4.4.0)
   utf8            1.2.4      2023-10-22 [1] CRAN (R 4.4.0)
   vctrs           0.6.5      2023-12-01 [1] CRAN (R 4.4.0)
   withr           3.0.2      2024-10-28 [1] CRAN (R 4.4.2)
   xfun            0.49       2024-10-31 [1] CRAN (R 4.4.2)
   yaml            2.3.10     2024-07-26 [1] CRAN (R 4.4.1)

 [1] C:/R/library
 [2] C:/R/R-4.4.2/library

 D ── DLL MD5 mismatch, broken installation.

```

</details>

Aariq:
In my quest to preserve metadata in raster targets I accidentally forgot to preserve the **data**.  We can't use `withr::local_tempdir()` in the read function because it'll get cleaned up and the resulting `SpatRaster` will have a broken file pointer.  I don't love this solution, but the only other workaround I can think of is to force the `SpatRaster` to point to memory (e.g. by adding 0 or multiplying by 1 after reading), and I like that solution less I think.

Fixes #127.

### DIFF
--- a/R/tar-terra-rast.R
+++ b/R/tar-terra-rast.R
@@ -137,7 +137,10 @@ tar_terra_rast <- function(name,
 tar_rast_read <- function(preserve_metadata) {
   switch(preserve_metadata,
     zip = function(path) {
-      tmp <- withr::local_tempdir()
+      tmp <- tempdir()
+      # NOTE: cannot use withr::local_tempdir() because the unzipped files need
+      # to persist so that the resulting `SpatRaster` object doesn't have a
+      # broken file pointer
       zip::unzip(zipfile = path, exdir = tmp)
       terra::rast(file.path(tmp, basename(path)))
     },
--- a/tests/testthat/test-tar-terra-rast.R
+++ b/tests/testthat/test-tar-terra-rast.R
@@ -143,11 +143,13 @@ tar_test("metadata is maintained", {
       r
     }
     list(
-      tar_terra_rast(r, make_rast())
+      tar_terra_rast(r, make_rast()),
+      tar_terra_rast(r2, make_rast(), preserve_metadata = "drop")
     )
   })
   tar_make()
   x <- tar_read(r)
   expect_equal(terra::units(x), rep("m", 3))
   expect_equal(terra::time(x), as.Date("2024-10-01") + c(0, 1, 2))
+  expect_equal(head(terra::values(x)), head(terra::values(tar_read(r2))))
 })


hansvancalster:
> I don't love this solution, but the only other workaround I can think of is to force the `SpatRaster` to point to memory (e.g. by adding 0 or multiplying by 1 after reading), and I like that solution less I think.

After reading https://github.com/ropensci/targets/discussions/809 again. I now ask myself whether `geotargets::tar_terra_read` would be better served by a combination of things that have been suggested there:

```r
...
tar_format(
  read = function(path) terra::unwrap(path),
  write = function(object, path) terra::wrapCache(object, filename =  path),
  marshal = function(object) terra::wrap(object),
  unmarshal = function(object) terra::unwrap(object)
)
...
```

So using wrap / unwrap both for the read/write and for marshal/unmarshal. I note that https://rspatial.github.io/terra/reference/wrapCache.html can be useful here instead of wrap(). 

Maybe this has already been looked at... It feels more elegant then the zip/unzip approach, but haven't tested this.

hansvancalster:
I have made a reprex that might be useful as a starting point for a more general solution for these problems with metadata.
The idea is to have a separate folder where geodata are stored next to the targets store. The command for geotargets::tar_terra_rast(command = ...) should write its output to the geodata store (which can be a .tif and associated .tif.aux.xml).  We can then use saveRDS(wrap(proxy = TRUE)) to write and unwrap(readRDS()) to read the target.  The object in the targetsstore will be small and contain no values, and the values will be recovered from geostore during unwrapping step.


``` r
make_rast <- function(geostore) {
  f <- system.file("ex/elev.tif", package = "terra")
  r <- terra::rast(f)
  r <- terra::classify(
    r, c(100, 300, 600),
    include.lowest = TRUE,
    brackets = TRUE,
    filename = file.path(geostore, "r2.tif")
    )
  return(r)
}

targetsstore <- "targetsstore"
geostore <- "geostore"

fs::dir_create(targetsstore)
fs::dir_create(geostore)

r <- make_rast(geostore = geostore)
r
#> class       : SpatRaster 
#> dimensions  : 90, 95, 1  (nrow, ncol, nlyr)
#> resolution  : 0.008333333, 0.008333333  (x, y)
#> extent      : 5.741667, 6.533333, 49.44167, 50.19167  (xmin, xmax, ymin, ymax)
#> coord. ref. : lon/lat WGS 84 (EPSG:4326) 
#> source      : r2.tif 
#> varname     : elev 
#> categories  : elevation 
#> name        : elevation 
#> min value   : [100–300] 
#> max value   : (300–600]

# write component of tar_custom
terra::wrap(r, proxy = TRUE) |>
  saveRDS(file = file.path(targetsstore, "rds_wrapped_r"))

# read component
w <- readRDS(file.path(targetsstore, "rds_wrapped_r"))
str(w)
#> Formal class 'PackedSpatRaster' [package "terra"] with 3 slots
#>   ..@ definition: chr "rast(ncols=95, nrows=90, nlyrs=1, xmin=5.74166666666667, xmax=6.53333333333333, ymin=49.4416666666667, ymax=50."| __truncated__
#>   ..@ values    : num[0 , 0 ] 
#>   ..@ attributes:List of 3
#>   .. ..$ sources :'data.frame':  1 obs. of  4 variables:
#>   .. .. ..$ sid   : int 1
#>   .. .. ..$ source: chr "C:/Users/hans_vancalster/AppData/Local/Temp/Rtmp4IQNXU/reprex-44f47e2b6e05-hemp-urson/geostore/r2.tif"
#>   .. .. ..$ bands : num 1
#>   .. .. ..$ nlyr  : num 1
#>   .. ..$ levels  :List of 1
#>   .. .. ..$ :'data.frame':   2 obs. of  2 variables:
#>   .. .. .. ..$ ID       : int [1:2] 0 1
#>   .. .. .. ..$ elevation: chr [1:2] "[100–300]" "(300–600]"
#>   .. ..$ levindex: int 1
w@values # no values
#> <0 x 0 matrix>
w@attributes$sources # points to geostore
#>   sid
#> 1   1
#>                                                                                                  source
#> 1 C:/Users/hans_vancalster/AppData/Local/Temp/Rtmp4IQNXU/reprex-44f47e2b6e05-hemp-urson/geostore/r2.tif
#>   bands nlyr
#> 1     1    1
rr <- terra::unwrap(w)
rr # the source is r2.tif in geostore, while we read this from targetsstore
#> class       : SpatRaster 
#> dimensions  : 90, 95, 1  (nrow, ncol, nlyr)
#> resolution  : 0.008333333, 0.008333333  (x, y)
#> extent      : 5.741667, 6.533333, 49.44167, 50.19167  (xmin, xmax, ymin, ymax)
#> coord. ref. : lon/lat WGS 84 (EPSG:4326) 
#> source      : r2.tif 
#> categories  : elevation 
#> name        : elevation 
#> min value   : [100–300] 
#> max value   : (300–600]
rr@ptr$filenames()
#> [1] "C:/Users/hans_vancalster/AppData/Local/Temp/Rtmp4IQNXU/reprex-44f47e2b6e05-hemp-urson/geostore/r2.tif"
```

<sup>Created on 2024-12-20 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Aariq:
@hansvancalster We've explored using wrapCache in the past (https://github.com/njtierney/geotargets/pull/63) and may reconsider it as an *option* in the future.

hansvancalster:
> @hansvancalster We've explored using wrapCache in the past (#63) and may reconsider it as an _option_ in the future.

Thanks! Happy to learn this has already been explored [and benchmarked](https://viz.datascience.arizona.edu/benchmarking-geotargets/benchmarking.html) 🚀  I agree with your assessment in the benchmark that either approach should be optional and I can certainly live with the current (i.e. this PR) implementation using zip-archive. I believe the "Cache" method might be preferable in case of large rasters (several Gb) because it avoids making a temporary copy of the raster (i.e. in the Cache method, the raster file is permanently stored and read from `_geotargets` cache-folder, whereas in zip-archive, the raster is zipped in `_targets` store-folder and needs to be unzipped to a temporary directory for the duration of the R-session, which requires temporarilly twice the disk space compared to Cache method).

Issue #132, opened 2024-12-23
Pass `datatype` to `writeRaster()` from `tar_terra_rast()`


brownag:
Currently we support passing `gdal` and `filetype` arguments to `writeRaster()`, but there are other arguments that are important or could be useful. These include datatype, NAflag, scale, offset, to name a few.

I think it is possible to pass additional arguments in `...`, which are not currently used in `tar_terra_rast()`, or to add the arguments that are most needed one at a time. We might just need to iterate over them and evaluate them to get constant values we can pass down through `tar_resources_custom_format()`

Perhaps the most important to add is `datatype` which matters when writing to file for geotargets use cases. `datatype` does not have a setter like NAflag, scale and offset, etc. do because SpatRaster objects in memory do not really have a concept of "data type" with the same granularity of the GDAL band data types unless  they have just been read from file or until they are written to file

The band type selected can dramatically affect the total size of target data files, read/write/compression/processing times, precision, etc.

@hansvancalster raised the following issue in https://github.com/njtierney/geotargets/issues/127#issuecomment-2550747264_ 
> - the datatype was changed. In the function get_map(), I used something like `m <- terra::writeRaster(..., datatype = "INT2U"); return(m)` to make things work nicely with raster files extracted from an ArcGis file geodatabase. Running this outside the pipeline returns the expected datatype, however in the pipeline it is changed to "INT4U".  This is probably a separate issue: allow passing `datatype` to `terra::writeRaster()` from `geotargets::tar_terra_rast()`. I also tried passing `gdal = c("DATATYPE = INT2U")` but that did not work. EDIT: I have removed the targets store and built everything again and I now get INT2U as datatype. Totally not clear to me why the problem dissapeared... EDIT2: the returned datatype seems to change unpredictably between different runs and may even differ between branches (it is always either INT4U or INT2U, whereas I would prefer it do be always INT2U for my use case).

The GDAL "DATATYPE" option does not exist, so users have no way to control the raster band type in geotargets, leading to unexpected results if they do something specific to their raster to set a more detailed data type beyond what the terra heuristics do.

The inconsistent behavior in @hansvancalster's case was due to a SpatRaster having been written to file manually within a target using `datatype = "INT2U"`, then written again using `tar_terra_rast()` which converts the integer data to `"INT4U"`. The datatype the user sees would depend on whether the object being checked was the one created by the previous target (preserving INT2U from manual write), or one loaded from the target store (preserving INT4U from target write). 

[...]

PR #137, opened 2025-01-09
pass additional arguments via `...` to writeRaster/writeVector/write_stars

brownag:
This is a possible fix for #132 (issues with `datatype` and additional arguments to `writeRaster()`) originally raised by @hansvancalster in #127. With this PR, additional arguments (`...`) passed to `tar_terra_rast()` and `tar_terra_vect()` are substituted into writeRaster/writeVector calls.

For example, with `tar_terra_rast(datatype= )`:
``` r
# write as FLT8S
targets::tar_script({
    library(targets)
    library(geotargets)
    list(
        
        tar_terra_rast(x, terra::rast(
            system.file("ex", "elev.tif", package = "terra")
        ), datatype = "FLT8S")
    
    )
})

targets::tar_make()
#> ▶ dispatched target x
#> ● completed target x [0.013 seconds, 13.106 kilobytes]
#> ▶ ended pipeline [0.174 seconds]
targets::tar_load(x)
terra::datatype(x)
#> [1] "FLT8S"

targets::tar_make()
#> ✔ skipped target x
#> ✔ skipped pipeline [0.088 seconds]
targets::tar_load(x)
terra::datatype(x)
#> [1] "FLT8S"

# changing datatype to INT2U invalidates the target
targets::tar_script({
    library(targets)
    library(geotargets)
    list(
        
        tar_terra_rast(x, terra::rast(
            system.file("ex", "elev.tif", package = "terra")
        ), datatype = "INT2U")
        
    )
})

targets::tar_make()
#> ▶ dispatched target x
#> ● completed target x [0.01 seconds, 8.4 kilobytes]
#> ▶ ended pipeline [0.221 seconds]
targets::tar_load(x)
terra::datatype(x)
#> [1] "INT2U"
```


Aariq:
Nice! I tried to figure out a way to get this to work without `do.call()` using `rlang::list2(...)` and `!!!args` instead (just for slightly easier readability), but that doesn't seem to play nicely with `tar_format()`. Should we keep `gdal` as an argument for `tar_terra_*()` or deprecate it (and the associated option) now that it gets passed via `...`?  Maybe fine to keep it since `gdal` seems like something you might want to set for an entire pipeline?

brownag:
> Should we keep `gdal` as an argument for `tar_terra_*()` or deprecate it (and the associated option) now that it gets passed via `...`? Maybe fine to keep it since `gdal` seems like something you might want to set for an entire pipeline?

I agree that the GDAL options and drivers may be more likely to be properties that we want to set for all or part of a pipeline, whereas `datatype` and friends can be more dataset/target specific. There will be cases where folks need specific option values for individual targets, and pipelines where folks want to handle all targets with the same specific settings, so we should make both easier if possible. 

IMO it makes sense to have `...` be passed to the "write" functions for targets, so I think this is a good use of the currently-unused `...` in geotargets functions.

I think it is fine to keep `gdal` and `options` as explicit arguments. As devil's advocate I would suggest `datatype` could be an explicit option because it affects target data size and precision, and also interacts with options such as those for compression. Other `writeRaster()` arguments are mostly used less frequently and have lower impact on the raster data itself. 

The ergonomics of `substitute` argument to `tar_format()` are pretty nice and do make me wonder whether we need all the extra options infrastructure that we developed before all these great additions to targets. But I think it is valuable to provide the option to use the options, as long as it doesn't get in the way of functionality or reproducibility.

R/tar-terra-rast.R
```diff
@@ -32,6 +32,20 @@
...
+#' @param preserve_metadata character. When `"drop"` (default), any
+#'   auxiliary files that would be written by [terra::writeRaster()] containing
+#'   raster metadata such as units and datetimes are lost (note that this does
+#'   not include layer names set with `names() <-`).  When `"zip"`, these
+#'   metadata are retained by archiving all written files as a zip file upon
+#'   writing and unzipping them upon reading. This adds extra overhead and will
+#'   slow pipelines. Also note metadata may be impacted by different versions
+#'   of GDAL and different drivers. If you have an issue with retaining
+#'   metadata for your setup, please file an issue at
+#'   \url{https://github.com/ropensci/geotargets/issues/} and we will try and
```
Aariq: This makes me think we should eventually create an issue template asking users to try a reprex with just `terra`.  We've seen a few issues opened in `geotargets` that are just `terra` behavior that we can't do much about.

Aariq (approved):
looks good to me.  I'm also OK with making `datatype` an explicit argument if you want to do that here or in a separate PR

R/tar-terra-rast.R
brownag: Agreed, an issue template would be good

brownag:
> looks good to me. I'm also OK with making `datatype` an explicit argument if you want to do that here or in a separate PR

I can do this very soon, in a separate PR, unless you want me to do it here

brownag:
Added a new geotargets option `"gdal.raster.data.type"` and implemented it for `tar_terra_rast()` and `tar_stars()`. 

I did this in this PR after all since we don't want to override/rewrite the internal handling of assigning default data type, which in terra at least is a relatively complex routine (as described in https://github.com/ropensci/geotargets/issues/132). 

As a workaround for the default case, the data type argument gets bolted on to the additional arguments list _only_ when specified/non-NULL. I only handled NULL for triggering the default but, I suppose we could have `NA` and `""` work equivalently (instead of letting the write methods error)

I also thought about creating a conversion function for the data type codes used in terra vs. stars: which would allow them to be used interchangeably for a pipeline that might use both packages and allows swapping a given option for the other. For now I just documented what is expected for each.

njtierney:
Thanks so much for this, @brownag and @Aariq ! 

Do we want to capture or test for if invalid data types are used? e.g.,

```r
Data type for writing raster file.
#'   One of: `"INT1U"`, `"INT2U"`, `"INT4U"`, `"INT8U"`, `"INT2S"`, `"INT4S"`,
#'   `"INT8S"`, `"FLT4S"`, `"FLT8S"` (for terra), or `"Byte"`, `"UInt16"`,
#'   `"UInt32"`, `"UInt64"`, `"Int16"`, `"Int32"`, `"Int64"`, `"Float32"`,
#'   `"Float64"` (for stars).
```

With `rlang::arg_match()` or similar? Or is it kind of unnecessary?

Also can you add a NEWS bullet?

brownag:
> Thanks so much for this, @brownag and @Aariq !
> 
> Do we want to capture or test for if invalid data types are used? e.g.,
> 
> ```r
> Data type for writing raster file.
> #'   One of: `"INT1U"`, `"INT2U"`, `"INT4U"`, `"INT8U"`, `"INT2S"`, `"INT4S"`,
> #'   `"INT8S"`, `"FLT4S"`, `"FLT8S"` (for terra), or `"Byte"`, `"UInt16"`,
> #'   `"UInt32"`, `"UInt64"`, `"Int16"`, `"Int32"`, `"Int64"`, `"Float32"`,
> #'   `"Float64"` (for stars).
> ```
> 
> With `rlang::arg_match()` or similar? Or is it kind of unnecessary?

I think both terra and stars will throw an error when given invalid value. Also, sometimes the allowable choices will depend on GDAL version available--i.e. INT1S / Int8 is relatively newly introduced (GDAL 3.7 ish?)in GDAL and I didn't put them in this list since they aren't in the terra/stars docs. Also some of the 64-bit variants aren't available on older GDAL etc. So we probably don't need any additional handling. I thought it would be useful to put the list of info, but perhaps should just link to the relevant docs in other package

> Also can you add a NEWS bullet?

Done in https://github.com/ropensci/geotargets/pull/137/commits/9819c268a1307f2109a3b70b2f7617380ea548df; reword as needed

njtierney:
> > Thanks so much for this, @brownag and @Aariq !
> > Do we want to capture or test for if invalid data types are used? e.g.,
> > ```r
> > Data type for writing raster file.
> > #'   One of: `"INT1U"`, `"INT2U"`, `"INT4U"`, `"INT8U"`, `"INT2S"`, `"INT4S"`,
> > #'   `"INT8S"`, `"FLT4S"`, `"FLT8S"` (for terra), or `"Byte"`, `"UInt16"`,
> > #'   `"UInt32"`, `"UInt64"`, `"Int16"`, `"Int32"`, `"Int64"`, `"Float32"`,
> > #'   `"Float64"` (for stars).
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > With `rlang::arg_match()` or similar? Or is it kind of unnecessary?
> 
> I think both terra and stars will throw an error when given invalid value. Also, sometimes the allowable choices will depend on GDAL version available--i.e. INT1S / Int8 is relatively newly introduced (GDAL 3.7 ish?)in GDAL and I didn't put them in this list since they aren't in the terra/stars docs. Also some of the 64-bit variants aren't available on older GDAL etc. So we probably don't need any additional handling. I thought it would be useful to put the list of info, but perhaps should just link to the relevant docs in other package
> 
> > Also can you add a NEWS bullet?
> 
> Done in [9819c26](https://github.com/ropensci/geotargets/commit/9819c268a1307f2109a3b70b2f7617380ea548df); reword as needed

Thanks for this! I think that all makes sense to me!

PR #140, opened 2025-01-10
add more tests

### LINKED ISSUE #139: Improve testing for test-tile-funs and tar_terra_sprc/sds
Based on comments from @amart90 in https://github.com/ropensci/software-review/issues/675

> * There could be additional testing (of course, there could always be more testing), particularly in the tile utility functions. Here are some examples I thought might be useful.   
>
>   * In `test-tile-funs.R` add tests for `tile_blocksize` and `tile_grid`. Additional tests could include that the expected number of tiles are being output and the sum of the number of cells is equal to the number of cells in the input raster.
>   * In the test `"recombined tiles are equal to original"`, it could be helpful to repeat the test for all 3 tile helper functions.
>   * For the tests `"tar_terra_sprc/sds() works"`, you have expectations for the class and snapshot. It might also be helpful to have tests to ensure some of the important metadata comes along, such as making sure the layer names, units, time, variables, and other attributes come along when expected. The snapshots for these tests do not include those metadata, so this may be as easy as assigning some of these attributes in the test SPRC/SDS and adding those fields in the snapshot.

njtierney:
- add checks and tests for integerish for tile_grid and tile_blocksize
- add a first pass test for tile_n returning the number of tiles being output

Resolves #139 

Still going with adding the tests!

### DIFF
--- a/tests/testthat/test-tar-terra-sprc.R
+++ b/tests/testthat/test-tar-terra-sprc.R
@@ -22,8 +22,6 @@ targets::tar_test("tar_terra_sprc() works", {
       # so this may be as easy as assigning some of these attributes in the
       # test SPRC/SDS and adding those fields in the snapshot.
       terra::units(rast_elev_scale) <- "m"
-      terra::varnames(rast_elev_scale) <- "elev"
-      terra::longnames(rast_elev_scale) <- "really-long-name"
       terra::time(rast_elev_scale) <- as.Date("2025-01-15")
       rast_elev_scale
     }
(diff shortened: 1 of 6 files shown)


tests/testthat/test-tar-terra-sprc.R
```diff
@@ -16,6 +16,16 @@ targets::tar_test("tar_terra_sprc() works", {
...
         projection
       )
+      # test to ensure metadata comes along.
+      # e.g., layer names, units, time, variables
+      # The snapshots for these tests do not include those metadata
+      # so this may be as easy as assigning some of these attributes in the
+      # test SPRC/SDS and adding those fields in the snapshot.
+      terra::units(rast_elev_scale) <- "m"
+      terra::varnames(rast_elev_scale) <- "elev"
+      terra::longnames(rast_elev_scale) <- "really-long-name"
```
Aariq, line 26: these won't be saved due to limitations in `terra`

njtierney: Got it, thanks!

PR #152, opened 2025-01-20
First pass at doing targets vignettes with just tar_script()

### LINKED ISSUE #148: Changing how targets code is presented in vignettes
From review over in https://github.com/ropensci/software-review/issues/675 - from @amart90:

> * Not all the code in the vignettes runs successfully. Specifically, code that the user should have populated in their _targets.R script files. Examples can be found in [Using `terra` with `geotargets`](http://geotargets.njtierney.com/articles/geotargets.html#tar_terra_rast-targets-with-terra-rasters) and [Example targets pipeline | Dynamic branching with raster tiles](http://geotargets.njtierney.com/articles/tar_terra_tiles.html#example-targets-pipeline). These sections do not not produce a working targets pipeline, so later calls to `tar_make()`, `tar_load()`, etc. throw errors. You call attention that this code should be the contents of the _targets.R file as a comment at the top of the code chunks; however, I think that this is probably too subtle and might trip up even casual targets users. At the very least, I think this deserves a bit more instruction to the user at the first instance in each vignette about copying the code into a _targets.R script file, possibly giving them the suggestion to paste this into `targets::tar_script()`. Another possibility would be to wrap the code in this section in `targets::tar_dir(targets::tar_script({ "code here" }))` the same way it is coded in your examples. This is not the easiest for users to see examples, but it has the benefit of not altering the user's global state (i.e., possibly overwriting an existing _targets.R script file). Once that was addressed, the code in both vignettes ran as expected, with the exception of the case noted above.

I agree with @aariq that we can leave it as is and we will then stay consistent with the rest of the targetopia packages, but I think I really like the idea of the user being able to copy any code chunk and it "just works". However, I think we should add a section at the start of the vignette that discusses how to run these examples - and then we should use `targets::tar_dir(targets::tar_script())` and then give an example of how that code would appear in a `_targets.R` file. I prefer this way because it means users can copy and paste the examples in the vignette easily, and we can refer the user back to that section on how to translate it into a targets pipeline/file.

njtierney:
Resolves #148

### DIFF
--- a/vignettes/geotargets.Rmd
+++ b/vignettes/geotargets.Rmd
@@ -43,11 +43,11 @@ In this vignette we will demonstrate the use of the `terra` R package, and we wi
 
 # How to run targets examples from vignettes
 
-The way targets typically works is you write a file named `_targets.R`, which describes the pipeline. See for example, the [`_targets.R` file in the demo-geotargets repo](https://github.com/njtierney/demo-geotargets/blob/main/_targets.R).
+The example code in this vignette is designed for you to be able to just copy and paste into the R console. However, this is not a typical way to run a targets workflow. 
 
-However, in order to demonstrate many of the features with `targets` and `geotargets`, we don't want to have to create many `_targets.R` files. So instead we use a targets function [`targets::tar_script()`](https://docs.ropensci.org/targets/reference/tar_script.html). This allows you to write the code you would have put in a `_targets.R` file. 
+The examples make use of [`targets::tar_script()`](https://docs.ropensci.org/targets/reference/tar_script.html), which creates or overwrites a `_targets.R` file (See for example, the [`_targets.R` file in the demo-geotargets repo](https://github.com/njtierney/demo-geotargets/blob/main/_targets.R)). In these examples, everything inside of `tar_script({})` is what would go inside a `_targets.R` file defining a workflow. When running the `tar_script` code, it will ask you each time if you want to overwrite the `_targets.R` file. This means if you are exploring these examples and copying the entire examples, it is worthwhile doing this in a separate repository to avoid overwriting your own `_targets.R` file.
 
-What this means for you is you can essentially just "copy and paste" the examples we provide in this vignette. When running the `tar_script` code, it will ask you each time if you want to overwrite the `_targets.R` file. This means if you are exploring these examples and copying the entire examples, it is worthwhile doing this in a separate repository to avoid overwriting your own `_targets.R` file.
+So, when building your own targets workflow it is **not recommended** to use `tar_script()`, but instead to edit `_targets.R` directly.
 
 ## `tar_terra_rast()`: targets with `terra` rasters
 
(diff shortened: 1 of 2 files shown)


njtierney:
I'd be curious on how this section looks that I added describing `tar_script` etc - what do you think, @Aariq ?

vignettes/geotargets.Rmd
```diff
@@ -41,15 +41,21 @@ In this vignette we will demonstrate the use of the `terra` R package, and we wi
...
 -   `tar_terra_sds()`
 
+# How to run targets examples from vignettes
+
+The way targets typically works is you write a file named `_targets.R`, which describes the pipeline. See for example, the [`_targets.R` file in the demo-geotargets repo](https://github.com/njtierney/demo-geotargets/blob/main/_targets.R).
+
+However, in order to demonstrate many of the features with `targets` and `geotargets`, we don't want to have to create many `_targets.R` files. So instead we use a targets function [`targets::tar_script()`](https://docs.ropensci.org/targets/reference/tar_script.html). This allows you to write the code you would have put in a `_targets.R` file. 
+
+What this means for you is you can essentially just "copy and paste" the examples we provide in this vignette. When running the `tar_script` code, it will ask you each time if you want to overwrite the `_targets.R` file. This means if you are exploring these examples and copying the entire examples, it is worthwhile doing this in a separate repository to avoid overwriting your own `_targets.R` file.
+
```
Aariq, line 51: I might flip the order here.  Something like:

>The example code in this vignette is designed for you to be able to just copy and paste into the R console.  However, this is not a typical way to run a `targets` workflow.  The examples make use of `targets::tar_script()` which creates or overwrites a `_targets.R` file.  In these examples, everything inside of `tar_script({})` is what would go inside a `_targets.R` file defining a workflow. When building your own `targets` workflow it is not recommended to use `tar_script()` but rather to edit `_targets.R` directly.

Aariq: Also, is it OK (e.g. according to CRAN) to use `tar_script()` without also wrapping it in `tar_dir()`?

vignettes/geotargets.Rmd
```diff
@@ -41,15 +41,21 @@ In this vignette we will demonstrate the use of the `terra` R package, and we wi
 -   `tar_terra_sprc()`
 -   `tar_terra_sds()`
 
+# How to run targets examples from vignettes
+
+The way targets typically works is you write a file named `_targets.R`, which describes the pipeline. See for example, the [`_targets.R` file in the demo-geotargets repo](https://github.com/njtierney/demo-geotargets/blob/main/_targets.R).
```
Aariq, line 46: Recommend using sentence breaks for easier diffs.  You can set this in RStudio options, for example.
```suggestion
The way targets typically works is you write a file named `_targets.R`, which describes the pipeline. 
See for example, the [`_targets.R` file in the demo-geotargets repo](https://github.com/njtierney/demo-geotargets/blob/main/_targets.R).
```

vignettes/geotargets.Rmd
njtierney: Not sure about the CRAN comment here, perhaps we shall have to find out!

vignettes/geotargets.Rmd
njtierney: I didn't know we could do this! I can't see how to change this option, I'll send you a message on slack.

PR #165, opened 2025-03-16
tar_stars: skip test with ncdf=TRUE when ncmeta is not installed

brownag:
This skips the `tar_stars()` tests that require use of `ncmeta` to write netCDF files when `ncmeta` is not installed.


njtierney (approved):
Looks good to me!

PR #169, opened 2025-05-01
Doc updates

brownag:
- add missing return value for `tar_terra_vrt()`
 - fix GDAL URLs that redirect in the source files. Some look like they are updated in .Rd, also fixed one in `tar_stars()`


Aariq (approved):

njtierney:
Thank you!

Issue #163, opened 2025-03-13
Release geotargets 0.3.0


njtierney:
First release:

* [x] `usethis::use_cran_comments()`
* [x] Update (aspirational) install instructions in README
* [x] Proofread `Title:` and `Description:`
* [x] Check that all exported functions have `@return` and `@examples`
* [x] Check [licensing of included files](https://r-pkgs.org/license.html#sec-code-you-bundle)
* [x] Review <https://github.com/DavisVaughan/extrachecks>

Prepare for release:

* [x] `git pull`
* [x] Check if any deprecation processes should be advanced, as described in [Gradual deprecation](https://lifecycle.r-lib.org/articles/communicate.html#gradual-deprecation)
* [x] `urlchecker::url_check()`
* [x] `devtools::build_readme()`
* [x] `devtools::check(remote = TRUE, manual = TRUE)`
* [x] `devtools::check_win_devel()`
* [x] `git push`
* [ ] Draft blog post
* [x] update CITATION.cff with `cffr::cff_write(dependencies = FALSE)` (after incrementing version)

Submit to CRAN:

* [x] `usethis::use_version('minor')`
* [x] `devtools::submit_cran()`
* [x] Approve email

Wait for CRAN...

* [x] Accepted :tada:
* [ ] Finish & publish blog post
* [ ] Add link to blog post in pkgdown news menu
* [x] `usethis::use_github_release()`
* [x] `usethis::use_dev_version(push = TRUE)`
* [ ] Share on social media

Aariq:
Should probably do `use_github_release()` and `use_dev_version()` before things get too messed up in `NEWS.md`.  E.g. in #138 the news item about adding `tar_terra_vrt()` should be under the development version heading, not under v0.3.0

njtierney:
Ah we actually got stalled in the CRAN review, need to do some minor fixes, so we should be able to put this into 0.3.0 :)

Aariq:
Anything I can do to help out with this?

njtierney:
Just submitted to CRAN now, fingers crossed it all goes through this time!

mdsumner:
I see the current report on CRAN about not having 3.7 GDAL, happy to help with that if you need. I came looking because I just found this: 

```
✔ ended pipeline [394ms, 8 completed, 0 skipped]
creating 1 * 1 = 1 tile extents
creating 1 * 2 = 2 tile extents
creating 1 * 3 = 3 tile extents
[...]

PR #172, opened 2025-05-12
Fix for CRAN error relating to GDAL version

Aariq:
Now throws an error if `preserve_metadata = "gdalraster_sozip"` and GDAL < 3.7.  Skips testing this feature when GDAL < 3.7 also.  This addresses the current [CRAN error](https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-clang/geotargets-00check.html).

TODO:
- Update NEWS.md after #163 is completed


brownag:
Thanks for catching this. Great to see it up on CRAN, sorry for not properly handling this version requirement in the tests from the beginning.

I think we should consider explicitly using `ubuntu-22.04` instead of `ubuntu-latest` platform so we can use an older GDAL in one of our R-CMD-check configs (perhaps just for oldrel). [Ubuntu 22.04 LTS uses GDAL 3.4.1](https://launchpad.net/ubuntu/jammy/+package/gdal-bin), I believe.

https://github.com/ropensci/geotargets/blob/25907bccc3d400adfe0c4a3476c277cba7f0f886/.github/workflows/R-CMD-check.yaml#L25

Aariq:
Good idea @brownag.  I can just add this to the matrix (instead of replacing ubuntu-latest + oldrel-1).  I'll add that to this PR

njtierney:
Thank you so much for this @Aariq  - I've just attached an issue for this fix, and I'll take a quick look at your PR later tonight.

njtierney:
My only thought from here is if we need to state System requirements like they do with terra:

https://github.com/USDAForestService/gdalraster/blob/main/DESCRIPTION#L74

@mdsumner do you have thoughts on this?

brownag:
> My only thought from here is if we need to state System requirements like they do with terra:
> 
> https://github.com/USDAForestService/gdalraster/blob/main/DESCRIPTION#L74
> 
> @mdsumner do you have thoughts on this?

Since we import terra it is implied that we have the same system requirements that compilation of terra depends on. I don't think we need to specify it in geotargets as we do not have any compiled code of our own that depends on those libraries. 

While it wouldn't be "wrong" to put it in there, SystemRequirements is just informational and does not actually directly affect any package configuration.

mdsumner:
that's right, it's not a SysReq for geotargets, you can  find the GDAL version for the package you're using `gdalraster::gdal_version()` - say. Just be aware that terra and gdalraster could report different versions depending on the history of installs, so whether you add extra messaging for users in case they have <3.7 (or whatever is relevant).

mdsumner:
see I have let vapour bitrot because gdalraster is so capable I haven't updated in a while (just to illustrate theoretically the versions could be different)

this on Windows, R 4.4.2

```R
terra::gdal()
[1] "3.10.1"
> gdalraster::gdal_version()
[1] "GDAL 3.10.1, released 2025/01/08" "3100100"                         
[3] "20250108"                         "3.10.1"                          
> 
> 
> vapour::vapour_gdal_version()
[1] "GDAL 3.9.3, released 2024/10/07"

```

brownag:
To @mdsumner's point, we probably do want to use @ctoney's approach from #172 in the optional gdalraster tests. 

https://github.com/ctoney/geotargets/blob/1f057528e740812b24d5bdf7b8f80ec1b45138a5/tests/testthat/test-tar-terra-rast.R#L204-L206

Since the libraries are statically "bundled" with the package install on Windows, rather than using the system's libs, the terra vs gdalraster version match is something that could come up pretty easily.

ctoney:
If you decide to do that, please note that #174 also changed DESCRIPTION to suggest `gdalraster (>= 2.0.0)`.

If you would rather not do that, then an alternative for the conditional that works with current 2.0 and previous gdalraster versions would be:
```r
skip_if(as.integer(gdalraster::gdal_version()[2]) < 3070000)
```

njtierney:
I'd like for users to be warned/errored if they aren't using GDAL version < 3.7.0, however it sounds like from above that we will need to depend on `gdalraster` in imports, which overall I'm OK with, but just worth noting.

mdsumner:
gdalraster can do so much now!  2.0.0 came out days ago, super impressive and massive work by @ctoney

ctoney:
An alternative could be to embed `terra::gdal(lib = "gdal")` version checking in your function `check_gdal_shz`, and embed `gdalraster::gdal_version()[4]` version checking in function `check_gdal_sozip`. It looks like your internal `gdal_version()` is only used in those two places currently (and one test) so it could be removed if you did that. This would let you avoid gdalraster as a hard dependency, and you can make the test for sozip conditional as in #174.

Please note that gdalraster has SysReq of GDAL >= 3.1.0, so this PR is effectively making that the minimum GDAL version for geotargets. That might be a good thing since 3.1 is pretty old now, and GDAL < 3.0 should probably be considered as not recommended for use at this point even though terra supports back to 2.2.3 theoretically. It also means that function `check_gdal_shz` with `min_version = "3.1"` will always pass now.

One other note, gdalraster depends on R ≥ 4.2.0, bumping up from R ≥ 4.1.0 that geotargets has currently.

Not meaning to impose any opinions here. I think the PR is fine as-is if you're comfortable with the implications. Just wanted to point those out, and a possible alternative if it raises concerns.

brownag:
Just to echo what @ctoney is saying--at this point I was envisioning only having gdalraster in Suggests, not Imports. 

We technically do not _need_ users to have both terra and gdalraster installed (except for the changes that were just made in https://github.com/ropensci/geotargets/pull/172/commits/7b904047ca91c4ee581a31da330b86198d38ab10). I personally think we don't need the `geotargets::gdal_version()` function as basically every R package that provides bindings to GDAL has a way to check the version. 

Furthermore, using gdalraster _everywhere_ for version checking means that we can't catch the potential mismatch issues that @mdsumner raised.

I could envision a future where the backend of geotargets always uses gdalraster--and other specific packages (terra, stars, sf, etc) are used only to materialize user-level objects. Currently terra serves that role as the main backend (and thus is in imports). I discussed this in a sortof long-winded way here: https://github.com/ropensci/geotargets/issues/4

njtierney:
Thanks for the discussion! Really appreciate this.

I have already submitted 0.3.1 to CRAN - but I think in the next release (unless this gets knocked back), I will roll back and remove the `Imports` dependency on `gdalraster`, since we aren't really using the features of the package (even though it is really great!). My apologies for rushing this, I am approaching a new contract that starts on Tuesday and wanted to get this submitted to CRAN as soon as I could to avoid the May 21 date of archival arriving. In hindsight I regret not fully finishing this discussion with you all before submitting to CRAN, because I really value your contributions to geotargets - without your help this package simply would not exist. In the future I will try to reach more of a consensus before CRAN submission.

Regarding:

> I think we don't need the `geotargets::gdal_version()` function as basically every R package that provides bindings to GDAL has a way to check the version. I suppose we could also add a package argument to gdal_version if we think that level of abstraction is necessary (and check the argument, and requireNamespace, as needed)

- The `geotargets:::gdal_version()` function is not exported, so it is not user facing.
- It uses `numeric_version()` which is a base function that simplifies testing package versions based on major.minor.patch, this simplifies use later when we are checking which version of a package people have installed.
- I wrote the function as it makes testing slightly easier using mocking, so that we can be sure that the error message does appropriately happen, but I think with recent changes in mocking I can get that behaviour with an external function.

> Furthermore, using gdalraster everywhere for version checking means that we can't catch the potential GDAL version mismatch issues that @mdsumner raised.

To explain my thinking, I thought that there was some difference between `terra::gdal(lib = "gdal")` and `gdalraster::gdal_version()`, and I guess there is, as it sounds like they can report/use different GDAL versions.

In our code for geotargets, especially for the relevant function, `tar_terra_rast()` it seems appropriate we use the `terra::gdal(lib = "gdal")` function to check GDAL version, is that right? My understanding is that this is the GDAL version that terra knows about/uses, is that correct?

> I could envision a future where the backend of geotargets always uses gdalraster--and other specific packages (terra, stars, sf, etc) are used only to materialize user-level objects. Currently terra serves that role as the main backend (and thus is in imports). In my original vision I was thinking all of the GDAL packages would actually be in suggests--thus we would not force stars users to install terra, terra users to install gdalraster, etc. unless their workflow needed it. I discussed this in a sort of long-winded way here: https://github.com/ropensci/geotargets/issues/4

I agree with you, as you said, that other specific pacakges (terra, stars, sf) are used only to materialise user-level objects, and that the GDAL packages could be in suggests, so users are only required to install the relevant packages that they need for their workflow.

I think it might be worthwhile to write up/move your thoughts on this idea of moving most GDAL packages into Suggests into a separate issue, combining what you've said here and in #4 - I'm happy to synthesise this into a new issue, if you like, @brownag 

Thank you again everyone for your discussion, again I want to emphasise that I want everyone to feel heard, I am sorry for rushing ahead with the 0.3.1 release, but I think we can move to a better spot from here.

brownag:
> > Furthermore, using gdalraster everywhere for version checking means that we can't catch the potential GDAL version mismatch issues that @mdsumner raised.
> 
> To explain my thinking, I thought that there was some difference between `terra::gdal(lib = "gdal")` and `gdalraster::gdal_version()`, and I guess there is, as it sounds like they can report/use different GDAL versions.
> 
> In our code for geotargets, especially for the relevant function, `tar_terra_rast()` it seems appropriate we use the `terra::gdal(lib = "gdal")` function to check GDAL version, is that right? My understanding is that this is the GDAL version that terra knows about/uses, is that correct?

Yes, that is correct.

> > I could envision a future where the backend of geotargets always uses gdalraster--and other specific packages (terra, stars, sf, etc) are used only to materialize user-level objects. Currently terra serves that role as the main backend (and thus is in imports). In my original vision I was thinking all of the GDAL packages would actually be in suggests--thus we would not force stars users to install terra, terra users to install gdalraster, etc. unless their workflow needed it. I discussed this in a sort of long-winded way here: #4
> 
> I agree with you, as you said, that other specific pacakges (terra, stars, sf) are used only to materialise user-level objects, and that the GDAL packages could be in suggests, so users are only required to install the relevant packages that they need for their workflow.
> 
> I think it might be worthwhile to write up/move your thoughts on this idea of moving most GDAL packages into Suggests into a separate issue, combining what you've said here and in #4 - I'm happy to synthesise this into a new issue, if you like, @brownag

I think we are ok at this time. The main thing we are working around is the serialization issue with terra objects, so it makes sense to have it the way we have it. 

As discussed in https://github.com/ropensci/geotargets/issues/48 I think we could apply a similar approach for gdalraster objects

To elaborate a bit more:  I envision a pathway to a more backend-agnostic geotargets for working with geospatial targets backed by formats written with GDAL. 

We can currently store an sf object, for instance, without serialization issues. So arguably a specific geotargets approach is not "needed" if regular targets works fine, but it still might be "better" for those data to be stored in the target store as a "real" geospatial format rather than binary representation of an R object. For instance, perhaps someday there will be breaking changes to how R serializes objects. I think still valuable to have specific methods to make working with the GDAL formats easier. Using a geospatial format also opens up the doors to "lazy" access or other types of database-like querying rather than having to materialize entire target objects in memory. I can write up an issue for it, but if you would like to capture it that would be fine with me too. :)

> Thank you again everyone for your discussion, again I want to emphasise that I want everyone to feel heard, I am sorry for rushing ahead with the 0.3.1 release, but I think we can move to a better spot from here.

Sorry if it came off like I was upset about this or feeling unheard. Just wanted to point out what I was thinking/had thought. Really appreciate all your work as maintainer here and shepherding all this work onto CRAN. I am so excited for more folks to be able to use it.

ctoney:
Appreciate your consideration of my comments and the opportunity for discussion here. Completely understandable to just get the CRAN update done for now and revisit in the next release. No issues feeling unheard.

njtierney:
Thanks everyone :)

I was confident that people weren't upset, however it is important to me everyone feels heard, and I'm happy you all did. I'd rather err on the side of checking in on this, because I want to prioritise this (continuing to be)ing a great space and project to work on.

Regarding the vision about serialisation and geospatial formats from @brownag, I really like this, because this:

>I think still valuable to have specific methods to make working with the GDAL formats easier. Using a geospatial format also opens up the doors to "lazy" access or other types of database-like querying rather than having to materialize entire target objects in memory. 

Sounds incredibly valuable to open the doors to "lazy" access.

> I can write up an issue for it, but if you would like to capture it that would be fine with me too. :)

When you have time, I'd love for you to write up an issue for this, that would be a great help @brownag 

From here, I will open an issue discussing moving `gdalraster` to Suggests, and rewriting some of the code around tests/checks for GDAL version etc.

PR #174, opened 2025-05-13
Fix test "metadata is maintained (gdalraster SOZip)" for GDAL < 3.7

ctoney:
This PR skips the test that uses `gdalraster::addFilesInZip()` if GDAL < 3.7.0. Currently showing as an error on CRAN at https://cran.r-project.org/web/checks/check_results_geotargets.html:

```
    ── Error ('test-tar-terra-rast.R:223:3'): metadata is maintained (gdalraster SOZip) ──
    <tar_condition_run/tar_condition_targets/rlang_error/rlang_error/error/condition>
    Error: Error in tar_make():
      _store_ addFilesInZip() requires GDAL >= 3.7
```

In case it's of interest, I also see a warning from this test that comes from a GDAL warning:

```
Warning (test-tar-terra-rast.R:228:3): metadata is maintained (gdalraster SOZip)
GDAL Message 1: NaN converted to INT_MAX.
```

Since that test uses an HFA file type, it may be due to: https://github.com/OSGeo/gdal/pull/12323

I'm using GDAL 3.10.3 on Ubuntu.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -47,7 +47,7 @@ Suggests:
     stars,
     testthat (>= 3.0.0),
     fs,
-    gdalraster,
+    gdalraster (>= 2.0.0),
     spelling
 Config/testthat/edition: 3
 URL: https://github.com/ropensci/geotargets, https://docs.ropensci.org/geotargets/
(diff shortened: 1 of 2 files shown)


ctoney:
Sorry, I had paused for a while to look at the warnings I was getting from the test set, and didn't check back to see you had this covered in the meantime. This can be ignored obviously. Nice work on the package.

njtierney:
Thanks for this, @ctoney - really appreciate you taking the time to submit a PR! @Aariq was onto this and I'm just tweaking a couple of things, we should be submitting to CRAN in the next day or so.

PR #185, opened 2025-10-03
Snapshot test update

Aariq:
This addresses failing pkgcheck by updating some snapshots that changed because of apparent changes in `terra` print methods.  It also updates documentation inherited from `targets`


njtierney (approved):
Looks good to me!

PR #186, opened 2025-10-07
add r-multiverse and cran badges

Aariq:
Adds badges to readme for CRAN status and r-multiverse status


njtierney (approved):
Looks good to me!

PR #189, opened 2026-04-01
fix(tests): remove snapshots of show method output

brownag:
Resolves failing tests from https://github.com/ropensci/geotargets/issues/177#issuecomment-4166847662


njtierney:
Thank you, @brownag !

PR #192, opened 2026-06-11
fix(tar_terra_*): skip some tests if crew not installed

brownag:
small fix for tests when suggested package crew isn't installed


Aariq (approved):
seems reasonable to me!

njtierney:
I agree with @Aariq ! Thank you, @brownag !

njtierney:
Oops, clicked "close" thinking it was "merge"

